PR #87, opened 2014-12-02
Txs single partition

angbrav:
Optimisation for transactions that affect only one partition. I have also added two tests to check that works. The operations clocksi_iprepare and clocks_icommit used for testing corner cases in the algorithm are still in the API. For benchmarking, the new operation clocks_full_icommit(we can change the name) should be used since it will take advantage of the optimisation and likely improve results.

### DIFF
--- a/riak_test/clocksi_test.erl
+++ b/riak_test/clocksi_test.erl
@@ -19,7 +19,7 @@
 %% -------------------------------------------------------------------
 -module(clocksi_test).
 
--export([confirm/0, clocksi_test1/1, clocksi_test2/1,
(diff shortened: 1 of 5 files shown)


bieniusa:
Manuel, this seems to be an optimization we still want to have, right?
Can you prepare it for merging into master?

tcrain:
These changes are included in https://github.com/SyncFree/antidote/pull/139

bieniusa:
Can this one be closed?

angbrav:
The changes are included in Tylers changes, which were already merged in master. I would say we can close it.

Manuel

PR #88, opened 2015-01-11
Added bounded counter CRDT.

lythq:
Also fixed error handling in transactions and downstream generation, as well as rename to Antidote in riak_test setup instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,12 +229,6 @@ Running Tests
 	* git checkout features/csm/floppystore
 2. cd RIAK_TEST and run commands
 	* make
-5. Download and put `.riak_test.config` in home directory from following link
-	https://gist.github.com/lythq/f979e11749a868fc33dd
-6. Create directory `rt` in home
-7. Modify  `.riak_test.config` 
-	* Change rtdev\_path (line number 106) to point to your HOME/rt/antidote
-	* Change test\_paths (line number 109) to point to your PATH\_TO\_ANTIDOTE/riak\_test/ebin
 
 ### Building antidote for testing ###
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -230,11 +230,11 @@ Running Tests
 2. cd RIAK_TEST and run commands
 	* make
 5. Download and put `.riak_test.config` in home directory from following link
-	https://gist.github.com/cmeiklejohn/d96d7167e1a8ed4b128d
+	https://gist.github.com/lythq/f979e11749a868fc33dd
```
cmeiklejohn: Care to clarify the changes between the two gists?

lythq: Yours still has the project name floppystore, but riak_test on Antidote is named antidote.
So you get an error if you haven't run riak_test on the previous code, and the new code isn't the one being tested.
I also added the test_paths setting, so you're less prone to skip that step.

cmeiklejohn: Great.  But, why even bother having it in a gist anymore?  How about just put the file in this repo?

lythq: Well pointed out. Automated that part and removed the gist.

bieniusa:
@cmeiklejohn I want to merge this pull request. Just one doubt: What about the Basho copyright notice? We are slowly moving from state-based to op-based. Would the copyright notice still apply?

lythq:
This CRDT isn't based on the ones from Riak, it was built from scratch only following the guidelines from other modules.

bieniusa:
Ok, let’s remove the copyright notice then.

bieniusa:
Just checked again. The pn_counter contains a copyright notice for basho - this one should go I think.

lythq:
pn_counter has nothing to do with this CRDT or this pull request, except for both being CRDTs and following the same guidelines...

PR #90, opened 2015-01-12
Fix build.

cmeiklejohn:
Resolve problems in the build where under a defined EQC compile, it
would attempt to export functions which were removed from the module.


cmeiklejohn:
Rebased and FORCE PUSHED.

PR #92, opened 2015-02-02
Fix some replication issues.

lythq:
Also added ring cluster name to replication info.

### DIFF
--- a/riak_test/append_test.erl
+++ b/riak_test/append_test.erl
@@ -32,6 +32,7 @@ confirm() ->
     rt:wait_until_ring_converged(Nodes),
 
     Node = hd(Nodes),
+    rt:wait_for_service(Node, antidote),
 
     rt:log_to_nodes(Nodes, "Starting write operation 1"),
 
--- a/riak_test/inter_dc_repl_test.erl
+++ b/riak_test/inter_dc_repl_test.erl
@@ -7,11 +7,18 @@
 -define(HARNESS, (rt_config:get(rt_harness))).
 
 confirm() ->
+    % Must be a power of 2, minimum 8 and maximum 1024.
+    rt:update_app_config(all,[
+        {riak_core, [{ring_creation_size, 8}]}
+    ]),
     [Cluster1, Cluster2] = rt:build_clusters([1,1]),
 
     Node1 = hd(Cluster1),
     Node2 = hd(Cluster2),
-    timer:sleep(100), %%TODO: wait for inter_dc_manager to be up
+
+    rt:wait_until_registered(Node1, inter_dc_manager),
+    rt:wait_until_registered(Node2, inter_dc_manager),
+
     {ok, DC1} = rpc:call(Node1, inter_dc_manager, start_receiver,[8091]),
     {ok, DC2} = rpc:call(Node2, inter_dc_manager, start_receiver,[8092]),
 
--- a/riak_test/multiple_dcs_test.erl
+++ b/riak_test/multiple_dcs_test.erl
@@ -7,6 +7,10 @@
 -define(HARNESS, (rt_config:get(rt_harness))).
 
 confirm() ->
+    % Must be a power of 2, minimum 8 and maximum 1024.
+    rt:update_app_config(all,[
+        {riak_core, [{ring_creation_size, 8}]}
+    ]),
     [Cluster1, Cluster2, Cluster3] = rt:build_clusters([1,1,1]),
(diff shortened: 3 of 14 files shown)


riak_test/append_test.erl
```diff
@@ -30,6 +30,7 @@ confirm() ->
 
     lager:info("Waiting for ring to converge."),
     rt:wait_until_ring_converged(Nodes),
+    timer:sleep(1000), %%TODO: wait for antidote to be up
```
cmeiklejohn: There is a `wait_until` function provided which I feel would be better used here.

riak_test/inter_dc_repl_test.erl
```diff
@@ -7,11 +7,15 @@
...
+    % Must be a power of 2, minimum 8 and maximum 1024.
+    rt:update_app_config(all,[
+        {riak_core, [{ring_creation_size, 8}]}
+    ]),
     [Cluster1, Cluster2] = rt:build_clusters([1,1]),
 
     Node1 = hd(Cluster1),
     Node2 = hd(Cluster2),
-    timer:sleep(100), %%TODO: wait for inter_dc_manager to be up
+    timer:sleep(6000), %%TODO: wait for inter_dc_manager to be up
```
cmeiklejohn: Similar comment.

riak_test/multiple_dcs_test.erl
```diff
@@ -15,7 +19,7 @@ confirm() ->
     rt:wait_until_ring_converged(Cluster1),
     rt:wait_until_ring_converged(Cluster2),
     rt:wait_until_ring_converged(Cluster3),
-    timer:sleep(500), %%TODO: wait for inter_dc_manager to be up
+    timer:sleep(6000), %%TODO: wait for inter_dc_manager to be up
```
cmeiklejohn: Similar comment.

riak_test/pb_client_test.erl
```diff
@@ -34,6 +34,7 @@ confirm() ->
 
     lager:info("Waiting for ring to converge."),
     rt:wait_until_ring_converged(Nodes),
+    timer:sleep(1000), %%TODO: wait for antidote to be up
```
cmeiklejohn: Similar comment.

src/inter_dc_repl_vnode.erl
```diff
@@ -64,42 +64,47 @@ init([Partition]) ->
...
+                    Transaction = {TxId, {DcId, Time}, Clock, Heartbeat},
+                    case inter_dc_communication_sender:propagate_sync(
+                           {replicate, [Transaction]}, DCs) of
+                        ok ->
+                            NewReader = NewReaderState;
+                        _ ->
+                            NewReader = NewReaderState
+                    end;
+                [_H|_T] ->
+                    case inter_dc_communication_sender:propagate_sync(
```
cmeiklejohn: It's better to do `Variable = case...` so you avoid warnings about exporting variables out of case statements.

lythq:
Thanks, those changes are now implemented.

cmeiklejohn:
These changes look good!  However, this test is still hanging for me: `multiple_dcs_test`.

bieniusa:
I also tried to run the test suite a couple of times. It was hanging each time with another test case, seemingly due to the riak_test setup.
What I find surprising is that these issues are so dominant with the changes in the branch. Does someone have a good explanation?

lythq:
Chris, can you please try it again? I believe the problem you were having is the auxiliary process which creates the socket listening to requests not being up. I changed the order of the delay in the replication vnode (yes, that should probably be changed as well), which is probably more than enough to solve that issue.

cmeiklejohn:
Still getting failures:

```
==> /Users/cmeiklejohn/rt/antidote/current/dev/dev3/log/console.log <==
2015-02-05 09:06:18.014 [error] <0.840.0>@inter_dc_communication_sender:connect:98 Couldnot connect to remote DC

==> /Users/cmeiklejohn/rt/antidote/current/dev/dev3/log/error.log <==
2015-02-05 09:06:18.014 [error] <0.840.0>@inter_dc_communication_sender:connect:98 Couldnot connect to remote DC
```

lythq:
This late commit, which makes the replication manager wait for the auxiliary process to create the TCP socket, seems to have solved Chris' error.
Annette can you please try it, and merge if successful?

bieniusa:
The direct synchronization from the last commit seems dubious to me. But I will pull it in such that we can make progress with the other stuff.

PR #99, opened 2015-05-04
Use supervisor pool.

cmeiklejohn:
Use a pool of supervisors to balance connections on; this eliminates the
serialization point at the supervisor for starting children, which was
always seen having a ~200ish backlog of messages under our profiling
workload.


cmeiklejohn:
summary

cmeiklejohn:
After pushing another change for the supervisor pools, the throughput has become much more stable.

cmeiklejohn:
summary

cmeiklejohn:
In the second one, you'll see throughout is more stable, and latency is much lower. (99th, 99.9th)

PR #100, opened 2015-05-12
Address incorrect function call.

cmeiklejohn:
antidote_pb_set.erl:70: Call to missing or unexported function lists:foreach/3

lists:foreach/3 doesn't exist; this call is supposed to technically be a
lists:foldl, however, it's unclear if we have any coverage for this
function at all, given none of the existing tests generated an error
with this call.


bieniusa:
Added issue to add test to increase code coverage.

PR #105, opened 2015-05-13
fixed {error, no_snapshot} bug in materializer_vnode

aletomsic:


bieniusa:
Is there a test that covers this case? If not, add one.

cmeiklejohn:
Fix looks good, however, for future reference we need to be a lot better with keeping this code a bit cleaner.  

For example:

screenshot 2015-05-15 11 44 11

cmeiklejohn:
This fix still doesn't handle the problem from the clients perspective, which needs to be addressed.

```
11:58:46.983 [error] CRASH REPORT Process <0.12369.0> with 0 neighbours exited with reason: {error,{badmatch,{error,no_snapshot}},[{antidote_pb_counter,process,2,[{file,"src/antidote_pb_counter.erl"},{line,74}]},{riak_api_pb_server,process_message,4,[{file,"src/riak_api_pb_server.erl"},{line,388}]},{riak_api_pb_server,connected,2,[{file,"src/riak_api_pb_server.erl"},{line,226}]},{riak_api_pb_server,decode_buffer,2,[{file,"src/riak_api_pb_server.erl"},{line,364}]},{gen_fsm,handle_msg,7,[{file,"gen_fsm.erl"},{line,505}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]} in gen_fsm:terminate/7 line 622
```

cmeiklejohn:
Opened https://github.com/SyncFree/antidote/issues/114 to address the client issue.

aletomsic:
This has to do with the fact that the protocol buffers client was no dealing with errors responses at all.
If I am not wrong, Li had solved this issue in some branch. 
Li, is this issue solved???

PR #110, opened 2015-05-14
Standalone static_tx_coord_fsm

marsleezm:
In the previous implementation, the operations of a transaction will firstly go to static_tx_coord_fsm and then be handled by interactive_tx_coord_fsm. This was designed for code reuse. However it is actually a performance bottleneck due to the cost of spawning extra fsm and communication cost between static and interactive fsm.
In this branch I integrated code from interactive_tx_coord_fsm to  static_tx_coord_fsm, so a static transaction is handled by static_tx_coord_fsm directly.
This has considerable performance improvement:
Throughput of append increases from about 3100 op/s to 3600 op/s.
Medium latency of append decreases from 6.3 ms to 5.6 ms.


riak_test/log_handoff_test.erl
```diff
@@ -75,6 +75,7 @@ test_handoff(RootNode, NewNode, NTestItems) ->
     %% added to the root node.  Note: systest_read() returns
     %% /non-matching/ items, so getting nothing back is good:
     lager:info("Validating data after handoff:"),
+    timer:sleep(100),
```
cmeiklejohn: No, not even once.

This will just not do, as it will ultimately result in a non-deterministic test.

src/clocksi_static_tx_coord_sup.erl
```diff
@@ -28,7 +28,7 @@
 
 -export([init/1]).
 
--define(NUM, 100).
+-define(NUM, 1).
```
cmeiklejohn: Why?  

This removes all of the pooling work I did to reduce contention on the supervisor and remove long message queues.

src/clocksi_static_tx_coord_fsm.erl
```diff
@@ -149,3 +250,87 @@ code_change(_OldVsn, StateName, State, _Extra) -> {ok, StateName, State}.
...
+wait_for_clock(DcId, Clock) ->
+   case get_snapshot_time(DcId) of
+       {ok, VecSnapshotTime} ->
+           case vectorclock:ge(VecSnapshotTime, Clock) of
+               true ->
+                   %% No need to wait
+                   {ok, VecSnapshotTime};
+               false ->
+                   %% wait for snapshot time to catch up with Client Clock
+                   timer:sleep(10),
```
cmeiklejohn: Why is this hardcoded at 10?

cmeiklejohn:
This branch generates significant type errors; please address.

```
==> antidote (dialyzer)
clocksi_interactive_tx_coord_fsm.erl:341: Type specification clocksi_interactive_tx_coord_fsm:get_snapshot_time(DcId::term(),ClientClock::vectorclock:vectorclock()) -> {'ok',vectorclock:vectorclock()} | {'error',term()} is a supertype of the success typing: clocksi_interactive_tx_coord_fsm:get_snapshot_time('undefined' | {_,_},dict()) -> {'error',_} | {'ok',dict()}
clocksi_interactive_tx_coord_fsm.erl:346: Type specification clocksi_interactive_tx_coord_fsm:get_snapshot_time(DcId::term()) -> {'ok',vectorclock:vectorclock()} | {'error',term()} is a supertype of the success typing: clocksi_interactive_tx_coord_fsm:get_snapshot_time('undefined' | {_,_}) -> {'error',_} | {'ok',dict()}
clocksi_static_tx_coord_fsm.erl:152: The call gen_fsm:reply(From::'undefined' | pid(),{'ok',_}) will never return since it differs in the 1st argument from the success typing arguments: ({_,_},any())
clocksi_static_tx_coord_fsm.erl:222: Expression produces a value of type {'error','commit_fail'} | {'ok',{'undefined' | #tx_id{server_pid::'undefined' | pid()},'undefined' | [any()],dict()}}, but this value is unmatched
clocksi_static_tx_coord_fsm.erl:263: Type specification clocksi_static_tx_coord_fsm:get_snapshot_time(DcId::term(),ClientClock::vectorclock:vectorclock()) -> {'ok',vectorclock:vectorclock()} | {'error',term()} is a supertype of the success typing: clocksi_static_tx_coord_fsm:get_snapshot_time('undefined' | {_,_},dict()) -> {'error',_} | {'ok',dict()}
clocksi_static_tx_coord_fsm.erl:303: Type specification clocksi_static_tx_coord_fsm:get_snapshot_time(DcId::term()) -> {'ok',vectorclock:vectorclock()} | {'error',term()} is a supertype of the success typing: clocksi_static_tx_coord_fsm:get_snapshot_time('undefined' | {_,_}) -> {'error',_} | {'ok',dict()}
logging_vnode.erl:178: Expression produces a value of type 'ok' | {'error','no_such_log' | 'nonode' | {'blocked_log',_} | {'read_only_mode',_} | {'file_error',string(),_}}, but this value is unmatched
make: *** [dialyzer] Error 1
```

riak_test/log_handoff_test.erl
marsleezm: The test passed on my local machine, but failed in Jenkins. So I guess that the problem is because vectorclock service is not started yet when the query to riak_core_apl:active_owners(vectorclock) is done, so I added this line to check and the test passed on Jenkins. Do you know any better way to solve this?

cmeiklejohn: "passed on my local machine" is not good enough.

You need to write some code that waits for the service to stop, and not rely on a hardcoded wait of a few milliseconds.

cmeiklejohn: There are plenty of examples of this in the code base already.

src/clocksi_static_tx_coord_sup.erl
marsleezm: For my machine, having one supervisor gives higher throughput than having 100 supervisors (3600 compared with 2800). Maybe the multiple supervisor setting will give more throughput on machines with more cores, but our targeting benchmark machine will only have two cores, so I think it is more appropriate to use just 1 here.

src/clocksi_static_tx_coord_fsm.erl
marsleezm: I guess just a reasonable value for spinning for clock. I copied from interactive_tx_coord

marsleezm:
I forgot to run dialyzer.. Will fix it soon.

src/clocksi_static_tx_coord_sup.erl
cmeiklejohn: Do you know why?  On the evaluation branch with my changes, the throughput when from under 4000 to 5000 with my changes.  I'm not going to merge this code until we know *why* that change is making something happen, given it's a pool of supervisors that it randomly balances on.  I want to know why this change makes a difference.

marsleezm:
I have problem running dialyzer.. Can you help? Thanks.

Checking whether the PLT .local_dialyzer_plt is up-to-date...
dialyzer: Not a regular file: /Users/liz/Documents/MyDocument/repositories/antidote/deps/ibrowse/ebin/ibrowse.beam

make: *** [.local_dialyzer_plt] Error 1

cmeiklejohn:
Remove ```.local_dialyzer_plt``` and have it reconstruct it.

src/clocksi_static_tx_coord_sup.erl
marsleezm: I ran several tests with different numbers of supervisors on the evaluation branch. The workload is append only, so it is a little lower than mixing read with append.
In the following result, the first number is the number of supervisor, the second is the throughput, the third is the average latency.
1 3275,15 3068,58
5 3094,34 3258,11
10 3103,03 3289,53
20 2882,41 3620,37
50 3109,1 3223,42
100 3081,78 3247,63

This is the result for my own static_coord branch.
1 3461,96 2915,79
10 3539,63 2838,74
50 3522,25 2833,74
100 3488,29 2856,16

cmeiklejohn: Again, as I've previously stated, I'm not going to merge a change that shows a number go up without knowing *why* the change is triggering that.  You still have not justified *why* the throughput is changing. 

Can you explain why changing: 

1 - 5: causes decrease in throughput
5 - 10: causes increase in throughput
10 - 20: decrease
etc.

We need to know what is really happening.

bieniusa:
I am a bit confused: This is a pull request for the evaluation branch, not for master.
And currently, it can't be automatically merged.

What should I do with it?

marsleezm:
I merged with evaluation branch. If you want to merge with master, I can also try to do it..

bieniusa:
Li and Manuel: What is the state of this branch? 
I didn't receive an answer why this was merged with the eval branch.

Again, what should I do with it?

tcrain:
The changes in this pull request should be covered in https://github.com/SyncFree/antidote/pull/139
But https://github.com/SyncFree/antidote/pull/139 is for master and this is for evaluation, so depending on what we want to use evaluation for can close or merge this

PR #115, opened 2015-05-15
Cleaning up the materializer

bieniusa:
Rewrote and restructured most parts of the materializer.
Starting to remove code duplication with clocksi_materializer.


cmeiklejohn:
Changes look good, however the test suite is failing.  I'm running again to see if the failure is just transient.

PR #124, opened 2015-05-26
Interdc fix

deepthidevaki:
A failure in sending heartbeats could result in a circular dependencies of metadata among DCs. This pull request has the fix for it and test to cover failure scenario.


src/inter_dc_repl_vnode.erl
```diff
@@ -91,8 +91,7 @@ handle_command(trigger, _Sender, State=#state{partition=Partition,
                 [_H|_T] ->
                     case inter_dc_communication_sender:propagate_sync(
                            {replicate, Transactions}, DCs) of
-                        ok -> NewReaderState;
-                        _  -> Reader
+                        ok -> NewReaderState
```
cmeiklejohn: Could the missing `_` case here generate a `case_clause` error and crash sometimes?

cmeiklejohn:
:+1:, seems good.

My only concern is about the non-exhaustive case clause I mentioned above, that could trigger a crash.

deepthidevaki:
the function returns always 'ok'. Dialyzer was issuing a warning with the additional clause '_'.

bieniusa:
Indeed, the error case can not happen anymore.

cmeiklejohn:
Not for nothing, but why have a `case` statement if it can only return one possible value?

PR #125, opened 2015-05-26
Fix abort message crashing interactive_coord_fsm

marsleezm:
An interactive txn that updates several keys can get aborted and receives several abort message.
In the current, as soon as it receives the first abort, it will transfer to reply_to_client state, as follows:
{next_state, reply_to_client, SD0#state{state=aborted},0};

So reply_to_client state is supposed to receive timeout message and start. However, it can receive other abort messages, which is not expected by the state, and the fsm will crash. So the fix is to add another event to trigger reply_to_client: if reply_to_client receives abort, it will proceed as if receives timeout.

### DIFF
--- a/src/clocksi_interactive_tx_coord_fsm.erl
+++ b/src/clocksi_interactive_tx_coord_fsm.erl
@@ -225,6 +225,8 @@ abort(abort, SD0=#state{transaction=Transaction,
 
 %% @doc when the transaction has committed or aborted,
 %%       a reply is sent to the client that started the transaction.
+reply_to_client(abort, SD) ->
+    reply_to_client(timeout, SD);
 reply_to_client(timeout, SD=#state{from=From, transaction=Transaction, state=TxState, commit_time=CommitTime}) ->
     case undefined =/= From of
         true ->


cmeiklejohn:
I'm confused.  

Shouldn't the client receive an `abort` message here and not a `timeout`?

marsleezm:
abort messages are already handled in previous states, like 
receive_prepared(abort, S0) -> ...
Then state transfer to abort, 
abort(timeout, SD0=#state{transaction=Transaction,
                          updated_partitions=UpdatedPartitions}) ->
    clocksi_vnode:abort(UpdatedPartitions, Transaction),
    {next_state, reply_to_client, SD0#state{state=aborted},0};
Finally it goes to reply_to_client, which expects to receive timeout as event, instead of an 'abort' again.

bieniusa:
I also don't understand this explanation.
Further, I keep getting timeouts at the beginning of the benchmark.
Then:
21:40:54.306 [warning] No data for op: {append,append}
21:40:54.307 [info] Errors:[{{append,append},98},{{{append,append},timeout},98}]
21:41:04.298 [warning] No data for op: {append,append}
And finally,
21:42:04.502 [debug] Driver basho_bench_driver_antidote_pb crashed: {{badmatch,{error,{tcp,etimedout}}},[{basho_bench_driver_antidote_pb,run,4,[{file,"src/basho_bench_driver_antidote_pb.erl"},{line,163}]},{basho_bench_worker,worker_next_op2,2,[{file,"src/basho_bench_worker.erl"},{line,236}]},{basho_bench_worker,worker_next_op,1,[{file,"src/basho_bench_worker.erl"},{line,242}]},{basho_bench_worker,max_worker_run_loop,1,[{file,"src/basho_bench_worker.erl"},{line,317}]}]}

bieniusa:
Can you, as we discussed already, provide also a test that shows first the failing behavior and then shows that it fixes the issue?

marsleezm:
Hi,
Firstly, this pull request is not intended to fix the timeout problem.

Secondly, the timeout on the master branch highly depends on the number of basho_bench benchmarking threads (concurrency). On my machine, the throughput of master is around 40 txns/sec, so if there are 40 threads, the average latency for a thread would be 1 second, and 2.5 seconds for 100 threads.. So I think the only thing we can do is to reduce the number of benchmarking threads. Using lots of threads will just overload the system and raise many timeouts.

Finally, there is another issue. Here I fixed the problem of txn_coord is crashed due to unexpected abort message, but basho_bench will still crash if I run the workload append_multiple. It's because this workload updates many keys so transactions may get aborted easily. However, the current pb data structure doesn't embeds failure reason, so it returns error for aborts transactions. And since error messages are too frequent, basho_bench gets crashed. Maybe @deepthidevaki  can help with this?

I will write the test.

marsleezm:
Hi, I could't think of a deterministic test case for this error, because this bug seems to depend on the message delivery order among multiple erlang processes. 
So let me explain this more concretely:
A coordinator C sends prepare to two partitions P1 and P2. Both P1 and P2 sends 'abort' back C. 
When C receives the first abort, it aborts the transaction and tells the result to the client, by transferring to the reply_to_client state, as the following command:
{next_state, reply_to_client, SD0#state{state=aborted},0};

So the next state is reply_to_client. After 0 ms, reply_to_client should receive an event 'timeout' and then start replying to clients.
reply_to_client(timeout, SD=#state{from=From, transaction=Transaction,....

However if we are unlucky, the second abort message can arrive before the timeout event, which is not expected by reply_to_client, and therefore will crash the fsm.
So my fix here is that, if reply_to_client receives abort message, it will just ignore it. That's all the fix does and it is not intended to solve the timeout issue. I don't think it is possible to write a deterministic test for it, because we can not control the order of message, unless I change the code to add some artificial delay.

If  anyone can suggest a way to do it, I would really appreciate your help.

bieniusa:
Regarding the test for the fsm, have a look at:
https://erlcode.wordpress.com/2010/09/05/eunit-testing-gen_fsm/ <https://erlcode.wordpress.com/2010/09/05/eunit-testing-gen_fsm/>

You will need to either implement or mock the partitions when testing the coordinator.

Questions:
- What do you mean with “reply_to_client” receives a timeout event? reply_to_client seems to be a state of the coordinator fsm.
You probably mean that the coordinator receives a timeout event (or abort event) while it is in the reply_to_client state.

- Your fix does not ignore the message, it instead triggers another timeout event. So, the coordinator will receive a timeout event for all partitions that want to trigger an abort. Since the fsm will stop after processing this first event, this behavior will create a lot of unnecessary messages etc. 
This is not what we want.

marsleezm:
Hi,

So for the first question, you are right.
For the second question, in my understanding, since fsm is a single-threaded process, only the first abort message or timeout will trigger reply_to_client. Then the fsm will enter terminate state and the remaining messages (abort or timeout) will just be ignored. Please correct me if I am wrong.

So for the test, to reproduce the error, we should make the fsm receive an abort in between its transition from 'abort' state and 'reply_to_client' state, before timeout is received. The only feasible way I can see is to increase the timeout value of 'abort' to 'reply_to_client' to some reasonable amount (like 1000ms) and to send abort message in-between. Without this change, even I mock each partitions, since the speed of state transaction of interactive_tx_coord_fsm and the speed to send abort message are arbitrary, it won't be easy to reproduce the error.

So the way I would suggest the following way to write the test:
1. Have some interleaving transactions that one transaction will be aborted and will receive several aborts from partitions.
2. Add random delay to clocksi_vnode (partitions) to send abort message, maybe between 10ms to 50ms (or some deterministic delay according to partition id), so that aborts will arrive one after each other and we need that to reproduce the error.
3. interactive_tx_coord_fsm has a delay (maybe 500ms) to transfer to reply_to_client, so it should receive abort message before receiving timeout.

What do you think? This seems to be a way with minimal modification to the code that can reproduce the error.

bieniusa:
> For the second question, in my understanding, since fsm is a single-threaded process, only the first abort message or timeout will trigger reply_to_client. Then the fsm will enter terminate state and the remaining messages (abort or timeout) will just be ignored. Please correct me if I am wrong.

Yes, this is what I was referring to: The fsm will terminate, and the remaining messages will not be processed. 
I couldn’t find any documentation though what happens to the messages.
To me it has a bad-code smell to have these messages sent, but not received.

> So for the test, to reproduce the error, we should make the fsm receive an abort in between its transition from 'abort' state and 'reply_to_client' state, before timeout is received. The only feasible way I can see is to increase the timeout value of 'abort' to 'reply_to_client' to some reasonable amount (like 1000ms) and to send abort message in-between. Without this change, even I mock each partitions, since the speed of state transaction of interactive_tx_coord_fsm and the speed to send abort message are arbitrary, it won't be easy to reproduce the error.

I don’t think I understand this proposal.
- Which process is sending the ‚timeout‘ message? If it is not the fsm itself, the test does not need to worry about timeouts - when unit testing, the test function is responsible for sending (and processing received) messages. 
Did you have a look at to the test setup for fsm's that I mentioned? https://erlcode.wordpress.com/2010/09/05/eunit-testing-gen_fsm/

- As you mentioned, fsms are single-threaded. It cannot receive a message between transitions, but maybe I do not understand what you mean.
The code for the abort state is the following. Which timeout value are you referring to?

abort(timeout, SD0=#state{transaction=Transaction,
                          updated_partitions=UpdatedPartitions}) ->
    clocksi_vnode:abort(UpdatedPartitions, Transaction),
    {next_state, reply_to_client, SD0#state{state=aborted},0};

abort(abort, SD0=#state{transaction=Transaction,
                        updated_partitions=UpdatedPartitions}) ->
    clocksi_vnode:abort(UpdatedPartitions, Transaction),
    {next_state, reply_to_client, SD0#state{state=aborted},0}.

> So the way I would suggest the following way to write the test:
> 1. Have some interleaving transactions that one transaction will be aborted and will receive several aborts from partitions.

No, write a simple eunit test suite for the fsm as explained the blog that I mentioned. This will cover all code paths in a simple way.

> 2. Add random delay to clocksi_vnode (partitions) to send abort message, maybe between 10ms to 50ms (or some deterministic delay according to partition id), so that aborts will arrive one after each other and we need that to reproduce the error.

This will still be non-deterministic and cause issues later.

> 3. interactive_tx_coord_fsm has a delay (maybe 500ms) to transfer to reply_to_client, so it should receive abort message before receiving timeout.

Code should not be adapted like this for regression tests.

	Annette

marsleezm:
A EUnit test won't work in this case, because interact_tx_coord_fsm needs to interact with the backend to proceed (unless I modify the code). To transition states (what we want is  init -> execute_op -> prepare -> abort -> reply_to_client ), interact_tx_coord_fsm needs to contact clocksi_vnode and perform update/read, which won't be possible if the backend(riak_core, clock_vnode etc) is not running. 
Overall, maybe I am missing something, but I don't see a way to write a deterministic test without modifying the code. Maybe it would be easier to discuss about it in the meeting.

bieniusa:
Li, what is the state of this branch? I think it has been resolved already by the recently added unit tests for the interact_txn_coord_fsm.

Can I close this issue?

marsleezm:
Yes, you are right! You can close it.

PR #128, opened 2015-06-06
Use "include_lib" instead of "include ../deps/"

essen:
Small tweak to make the code compile regardless of current working directory.


bieniusa:
Thank you!

PR #132, opened 2015-06-10
Add mock partition fsm and add macro to clocksi_interactive_coord_fsm

marsleezm:
I added some macro definition to clocksi_interactive_coord_fsm so that some components can be easily replaced. This enables testing the state transition of clocksi_interactive_coord_fsm by EUnit test in a light way.

As for bug fix, I changed reply_to_client from a state to a function. I also found a bug that, in abort state, the fsm may receive prepare message, which is not expected, and that will crash the fsm. Now the message is handled in abort state.

### DIFF
--- a/src/clocksi_interactive_tx_coord_fsm.erl
+++ b/src/clocksi_interactive_tx_coord_fsm.erl
@@ -342,6 +342,7 @@ main_test_() ->
      [
       fun empty_prepare_test/1,
       fun timeout_test/1,
+
       fun update_single_abort_test/1,
       fun update_single_success_test/1,
       fun update_multi_abort_test1/1,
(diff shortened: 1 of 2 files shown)


bieniusa:
Ah, this starts to look very nice!!

What still needs to be fixed is this dialyzer error:
clocksi_interactive_tx_coord_fsm.erl:262: The variable Reason can never match since previous clauses completely covered the type 'aborted' | 'committed'

If you have a look at the coverage for the eunit tests, there are some functions not covered yet (but much better than before!!). Can you try to increase coverage a bit more (by covering e.g. wait_for_clock)?

marsleezm:
A few more tests are added and the dialyzer warning is fixed. Please check it out.

bieniusa:
Again, much better!
Can we also get tests for the commit and receive_commit?

PR #134, opened 2015-06-16
Small fixes in inter_dc update propagation

tcrain:
This branch has a few fixes to the inter_dc update propagation.

First instead of creating an fsm directly in the sender, it uses a supervisor.  This way it doesn't have to use an fsm timeout on top of a connection timeout, because now if the fsm crashes the supervisor will restart it.

Also  on line 68 of src/inter_dc_communication_sender.erl an incorrect structure for the tuple was being used which caused a crash on line 57 in case of a timeout.  This is no longer needed because the supervisor is used.

Also slowed down the replication period/connection timeout, which before was creating cascading timeouts, this should make the tests more stable.  It is not a solution though, because the replication period should be much smaller in the first place, but since the inter dc replication system will be changed to use pub/sub this should be suitable for now.

### DIFF
--- a/src/inter_dc_communication_sender.erl
+++ b/src/inter_dc_communication_sender.erl
@@ -117,10 +117,12 @@ wait_for_ack(acknowledge, State)->
 wait_for_ack(timeout, State) ->
     %%TODO: Retry if needed
     lager:error("Timeout in wait for ACK",[]),
+    %% Retry after timeout is handled in the propagate_sync loop above
+    %% So the fsm returns a normal stop so that it isn't restarted by the supervisor
     {next_state,stop_error,State#state{reason=timeout},0}.
 
 stop(timeout, State=#state{socket=Socket}) ->
-    _ = gen_tcp:close(Socket),
+    _ = gen_tcp:close(Socket), 
     {stop, normal, State}.
 
 stop_error(timeout, State=#state{socket=Socket}) ->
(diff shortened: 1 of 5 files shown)


src/inter_dc_communication_sender.erl
```diff
@@ -132,7 +125,7 @@ stop(timeout, State=#state{socket=Socket}) ->
 
 stop_error(timeout, State=#state{socket=Socket}) ->
     _ = gen_tcp:close(Socket),
-    {stop, error, State}.
+    {stop, normal, State}.
```
cmeiklejohn: This will cause the process to not be restarted in the event of a timeout, is that correct behavior?

cmeiklejohn:
One general comment regarding the timeout behavior, otherwise :+1: to merge.

src/inter_dc_communication_sender.erl
tcrain: Yes because there is code to restart it in the propagate sync function. So before there were two places checking for timeouts.

Actually this is a question I have for the code in general. Is there a general rule for when is best for using the supervisor to  restart and when to do it manually. Bad messages? Connection timeouts? Receive timeouts? Etc.. Or is it just decided depending on the situation?

cmeiklejohn: I would probably leave a comment so that's clear to the reader.  I feel that you could go either way -- we do a lot of restarting due to connection timeouts via the supervisor by returning non-normal exits, but you don't have to do it that way.  As long as it's documented and clear.

cmeiklejohn:
:+1: to merge

PR #135, opened 2015-06-16
Have prepare block per key

tcrain:
The main purpose of this change is so that transactions in the prepare phase only block transactions trying to read the keys being prepared, before reads to any key in the partition would be blocked. In small tests this probably doesn't make a difference (it might even be slightly slower for certain workloads), but in larger systems with network delays and large transactions this could allow for much more concurrency.

This change also fixes an issue where a prepared transaction in progress would get overwritten by a new transaction being prepared with a larger prepare time, which could violate transaction atomicity if there is a concurrent read.

Added a simple test to check a read is blocked when there is a concurrent transaction preparing a read.

### DIFF
--- a/src/clocksi_vnode.erl
+++ b/src/clocksi_vnode.erl
@@ -354,9 +354,19 @@ clean_and_notify(TxId, Updates, #state{active_txs_per_key=_ActiveTxsPerKey,
 clean_prepared(_PreparedTx,[],_TxId) ->
     ok;
 clean_prepared(PreparedTx,[{_, {Key, _Type, {_Op, _Actor}}} | Rest],TxId) ->
-    [{Key,ActiveTxs}] = ets:lookup(PreparedTx, Key),
+    ActiveTxs = case ets:lookup(PreparedTx, Key) of
+		    [] ->
+			[];
+		    [{Key,List}] ->
+			List
+		end,
     NewActive = lists:keydelete(TxId,1,ActiveTxs),
-    true = ets:insert(PreparedTx, {Key, NewActive}),
+    true = case NewActive of
+	       [] ->
+		   ets:delete(PreparedTx, Key);
+	       _ ->
+		   ets:insert(PreparedTx, {Key, NewActive})
+	   end,
     clean_prepared(PreparedTx,Rest,TxId).
 
 
--- a/src/clocksi_vnode.erl
+++ b/src/clocksi_vnode.erl
@@ -288,7 +288,7 @@ handle_command({get_active_txns}, _Sender,
 						[] ->
 						    Acc;
 						_ ->
-						    [List|Acc]
+						    List ++ Acc
 					    end
 				    end,
     				    [],[{Key1,List1}|Rest1])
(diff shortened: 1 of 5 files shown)


src/clocksi_vnode.erl
```diff
@@ -125,13 +125,13 @@ abort(ListofNodes, TxId) ->
...
-                         [set, {write_concurrency, true}]),
+                         [set]),
     CommittedTx = ets:new(list_to_atom(atom_to_list(committed_tx) ++
                                            integer_to_list(Partition)),
-                          [set, {write_concurrency, true}]),
+                          [set]),
     ActiveTxsPerKey = ets:new(list_to_atom(atom_to_list(active_txs_per_key)
                                            ++ integer_to_list(Partition)),
-                              [bag, {write_concurrency, true}]),
+                              [bag]),
```
cmeiklejohn: Why the removal of the `write_concurrency` option?

src/clocksi_vnode.erl
```diff
@@ -125,13 +125,13 @@ abort(ListofNodes, TxId) ->
 init([Partition]) ->
     PreparedTx = ets:new(list_to_atom(atom_to_list(prepared_tx) ++
                                           integer_to_list(Partition)),
-                         [set, {write_concurrency, true}]),
+                         [set]),
```
cmeiklejohn: Why the removal of the `write_concurrency` option?

src/clocksi_vnode.erl
```diff
@@ -125,13 +125,13 @@ abort(ListofNodes, TxId) ->
...
     CommittedTx = ets:new(list_to_atom(atom_to_list(committed_tx) ++
                                            integer_to_list(Partition)),
-                          [set, {write_concurrency, true}]),
+                          [set]),
```
cmeiklejohn: Why the removal of the `write_concurrency` option?

cmeiklejohn:
This branch stalls the inter_dc_repl test indefinitely.

Errors like the following appear in the log:

```
2015-06-16 14:00:35.378 [error] <0.646.0> CRASH REPORT Process <0.646.0> with 0 neighbours exited with reason: {timeout,{gen_server,call,[{clocksi_vnode_master,'dev2@127.0.0.1'},{riak_vnode_req_v1,1278813932664540053428224228626747642198940975104,{server,undefined,undefined},{get_active_txns}},120000]}} in gen_fsm:terminate/7 line 622
2015-06-16 14:00:35.378 [error] <0.127.0> Supervisor riak_core_vnode_sup had child undefined started with {riak_core_vnode,start_link,undefined} at <0.646.0> exit with reason {timeout,{gen_server,call,[{clocksi_vnode_master,'dev2@127.0.0.1'},{riak_vnode_req_v1,1278813932664540053428224228626747642198940975104,{server,undefined,undefined},{get_active_txns}},120000]}} in context child_terminated
```

src/clocksi_vnode.erl
tcrain: From the docs:
{write_concurrency,boolean()} Performance tuning. Default is false, in which case an operation that mutates (writes to) the table will obtain exclusive access, blocking any concurrent access of the same table until finished. If set to true, the table is optimized towards concurrent write access. Different objects of the same table can be mutated (and read) by concurrent processes. This is achieved to some degree at the expense of memory consumption and the performance of sequential access and concurrent reading.

These tables are accessed only by this vnode which is single threaded.  Having write concurrency set to true makes it confusing because it seems like it should be accessed by several concurrent writers.  Also it means more memory consumption and slower access.

cmeiklejohn: That'd be a good thing to mention in the commit message, otherwise someone may think that they previously were being access by multiple processes. :)

tcrain:
Yep, thats broken, I didn't split my branches correctly.  I'll fix it tomorrow.

src/clocksi_vnode.erl
tcrain: Ok true, I will try to be more precise with commit messages in the future.

cmeiklejohn:
Sounds good; let me know when I should re-review.

cmeiklejohn:
Can you merge the latest master into this branch?

cmeiklejohn:
:+1: to merge

PR #139, opened 2015-06-26
Materialiser fixes/optimisations + read/write path optimisations + merge with evaluation branch

tcrain:
Sorry for the size of this pull request, it is a result of merging with another largely refactored branch, future ones will be smaller.  If there are certain changes that you feel are bad or buggy or not necessary, let me know and I will remove them.  These changes are broken down into three main components, described in the following:

1: Materialiser fixes/optimisations: This update fixes an issue where in certain cases garbage collection of old versions would cause updates to be lost, as well as issues where snapshots could be generated incorrectly when multiple DCs were performing updates.  Unit tests are added to check this.  Optimisations were done to allow for different snapshots to be materialised concurrently as well as storing the snapshots and operations ordered so that only enough of them are processed to be safe when generating new snapshots.  An new dictionary data structure is created for this that (partially) orders elements by vector clocks.  New snapshots are only stored if they are more recent than what is previously available, the reasoning behind this is that new transactions are started with the most recent available snapshot, so old snapshots probably won't be accessed often, though of course this depends on the workload.

2: Read/write path optimisations: The goal of these changes is to avoid unnecessary serialisation points and bottlenecks.  For example currently a read goes through the transaction coordinator, and then is serialised at the clocksi vnode where a read fsm is created, and then is serialised again at the materialiser vnode where a snapshot is generated.  This branch changes that so the transaction coordinator sends a request to a read server running at a vnode which directly processes the updates from the ets tables.  Each vnode has a configurable number of concurrent read servers.  This can be especially helpful for keys that have really busy read workloads.  Of course this also means we have to be careful with updates being prepared concurrently, requiring the prepared table to be written to twice during commit to ensure transaction atomicity.  Single read operations send their request directly to the read server instead of first starting up a transaction coordinator. Because of these changes, before performing reads the system needs to check that the vnodes have been initialised, the check_ready functions provide this functionality. 

Concerning writes, they are now cached in the transaction coordinator then sent to clocksi vnode during prepare, removing another point of serialisation.  Logging the updates happens during the prepare phase, though I might suggest moving the logging back to when the updated is executed in the transaction since it can be done asynchronously.  There are still many more optimisations to be done such as blind writes and caching reads, but these can be done in future pull requests.

3: Merging with evaluation branch:  This branch is merged into the evaluation branch, which included many changes from other people.  I think the main changes in this concern improvements to the transaction coordinator.   For example, the coordinator now chooses what method a transaction will use to commit based on the number of sites it updates.  For example if the transaction only updates a single site the prepare and commit can happen in a single round.  To keep with the current interface, the client still has to first call prepare then commit.  This branch also changes the logging vnode so that is uses erlang disk log instead of dets.

This branch still uses the old validation check for ensuring per dc snapshot isolation. Li has created a new version of this in a separate branch which will be a later pull request.

PS I cannot get the pb_client_test to run, due to some dependency error, maybe someone else has the same problem?


riak_test/wait_init.erl
```diff
@@ -0,0 +1,40 @@
+-module(wait_init).
```
bieniusa: This file misses comments, copyright, type annotations, tests.....

src/antidote.erl
```diff
@@ -132,8 +128,22 @@ clocksi_iread({_, _, CoordFsmPid}, Key, Type) ->
 clocksi_iupdate({_, _, CoordFsmPid}, Key, Type, OpParams) ->
     gen_fsm:sync_send_event(CoordFsmPid, {update, {Key, Type, OpParams}}).
 
+%% @doc This commits includes both prepare and commit phase. Thus
+%%      Client do not need to send to message to complete the 2PC
+%%      protocol. The Tx coordinator will pick the best strategie
+%%      automatically.
+%%      To keep with the current api this is still done in 2 steps,
```
bieniusa: Please add type annotation!

src/antidote_sup.erl
```diff
@@ -81,7 +81,12 @@ init(_Args) ->
                             {clocksi_interactive_tx_coord_sup, start_link, []},
                             permanent, 5000, supervisor,
                             [clockSI_interactive_tx_coord_sup]},
-
+    
+    ClockSiReadSup = {clocksi_readitem_sup,
+    		      {clocksi_readitem_sup, start_link, []},
```
bieniusa: Please adapt to naming convention, it probably should be ClockSIReadSup

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -49,43 +49,38 @@
...
 
 %% Callbacks
--export([init/1,
-         stop/1,
-         code_change/4,
-         handle_event/3,
-         handle_info/3,
-         handle_sync_event/4,
-         terminate/3]).
+-export([init/1, code_change/4, handle_event/3, handle_info/3,
```
bieniusa: Fix indenting, every function = one line.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -49,43 +49,38 @@
...
 
 %% States
--export([execute_op/3,
-         finish_op/3,
-         prepare/2,
-         receive_prepared/2,
-         committing/3,
-         receive_committed/2,
-         abort/2]).
+-export([execute_op/3, finish_op/3, prepare/2, prepare_2pc/2,
```
bieniusa: Fix indenting: Every function = one line

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -108,92 +103,176 @@ stop(Pid) -> gen_fsm:sync_send_all_state_event(Pid,stop).
...
+    %% Seed the random because you pick a random read server, this is stored in the process state
+    _Res = random:seed(now()),
     {ok, SnapshotTime} = case ClientClock of
-        ignore ->
-            get_snapshot_time();
-        _ ->
-            get_snapshot_time(ClientClock)
-    end,
+			     ignore ->
+				 get_snapshot_time();
```
bieniusa: Fix indenting!

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -108,92 +103,176 @@ stop(Pid) -> gen_fsm:sync_send_all_state_event(Pid,stop).
...
+
 
 %% @doc Contact the leader computed in the prepare state for it to execute the
 %%      operation, wait for it to finish (synchronous) and go to the prepareOP
 %%       to execute the next operation.
-execute_op({OpType, Args}, Sender,
-           SD0=#state{transaction=Transaction, updated_partitions=UpdatedPartitions}) ->
-    case OpType of
+execute_op({Op_type, Args}, Sender,
+           SD0=#state{transaction=Transaction,
```
bieniusa: Fix Naming convention: Variable names are camel case.

riak_test/clocksi_test.erl
```diff
@@ -19,27 +19,39 @@
...
+    clocksi_test3(Nodes),
     clocksi_test_prepare(Nodes),
+    clocksi_test5(Nodes),
     clocksi_tx_noclock_test(Nodes),
     clocksi_single_key_update_read_test(Nodes),
     clocksi_multiple_key_update_read_test(Nodes),
     clocksi_test4 (Nodes),
     clocksi_test_read_time(Nodes),
     clocksi_test_read_wait(Nodes),
-    clocksi_test_certification_check(Nodes),
```
bieniusa: Remove the trailing whitespace.

bieniusa:
Tyler, can you check with Manuel how the changes to the riak_tests relate to his refactorings?

src/clocksi_readitem_fsm.erl
```diff
@@ -28,110 +28,233 @@
...
--export([start_link/6]).
+-export([start_link/2]).
 
 %% Callbacks
 -export([init/1,
-         code_change/4,
+	 handle_call/3,
+	 handle_cast/2,
+         code_change/3,
          handle_event/3,
```
bieniusa: Fix indenting.

src/clocksi_readitem_fsm.erl
```diff
@@ -28,110 +28,233 @@
...
+    end.
+
+%% @doc This checks all partitions in the system to see if all read
+%%      servers have been started up.
+%%      Returns true if they have been, false otherwise.
+check_servers_ready() ->
+    {ok, CHBin} = riak_core_ring_manager:get_chash_bin(),
+    PartitionList = chashbin:to_list(CHBin),
+    check_server_ready(PartitionList).
+
```
bieniusa: Add type signatures.

src/clocksi_vnode.erl
```diff
@@ -18,17 +18,17 @@
...
 -behaviour(riak_core_vnode).
 
 -include("antidote.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
 -export([start_vnode/1,
-         read_data_item/4,
-         update_data_item/5,
+	 read_data_item/5,
+	 get_cache_name/2,
```
bieniusa: Fix indenting.

src/logging_vnode.erl
```diff
@@ -75,9 +77,16 @@ asyn_read_from(Preflist, Log, From) ->
...
-    riak_core_vnode_master:sync_command(Node,
+    case riak_core_vnode_master:sync_command(Node,
                                         {read_from, LogId, From},
-                                        ?LOGGING_MASTER).
+                                        ?LOGGING_MASTER) of
+        {ok, []} ->
+            {ok, []};
+        {ok, [H|T]} ->
+            {ok, [H|T]};
+        {error, Reason} ->
```
bieniusa: This can be simplified, there is no need to pattern match here, or is there one?

src/logging_vnode.erl
```diff
@@ -115,7 +131,14 @@ init([Partition]) ->
     LogFile = integer_to_list(Partition),
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     GrossPreflists = riak_core_ring:all_preflists(Ring, ?N),
-    Preflists = lists:filter(fun(X) -> preflist_member(Partition, X) end, GrossPreflists),
+    Preflists = lists:foldl(fun(X, Filtered) ->
+                                    case preflist_member(Partition, X) of
```
bieniusa: This change seems wrong: Filter should be more performant than fold (O(N) vs O(Nˆ2)) due to the append to the list.

src/logging_vnode.erl
```diff
@@ -176,27 +207,18 @@ handle_command({read_from, LogId, From}, _Sender,
 %%              OpId: Unique operation id
 %%      Output: {ok, {vnode_id, op_id}} | {error, Reason}
 %%
-handle_command({append, LogId, Payload}, Sender,
+handle_command({append, LogId, Payload}, _Sender,
                #state{logs_map=Map,
                       clock=Clock,
                       partition=Partition,
-                      senders_awaiting_ack=SendersAwaitingAck0}=State) ->
+                      senders_awaiting_ack=_SendersAwaitingAck0}=State) ->
```
bieniusa: If you don't need senders_awaiting_ack, you don't need to pattern match on it.

src/materializer_vnode.erl
```diff
@@ -35,8 +35,12 @@
 
 %% API
 -export([start_vnode/1,
-         read/4,
-         update/2]).
+	 check_tables_ready/0,
+         read/6,
+	 get_cache_name/2,
```
bieniusa: Fix indenting.

src/vector_orddict.erl
```diff
@@ -0,0 +1,129 @@
+%% -------------------------------------------------------------------
+%%
```
bieniusa: What about tests?

bieniusa:
For me, all (riak-)tests run. But make rial-tests end with an error, because it tries to test the wait_init.

tcrain:
Some of the fixes were for the type signatures so I went through the code and added or made more precise many of the dialyzer definitions.

While doing this I noticed that the structure of an operation record was different depending on whether it was part of a static or interactive transaction.  For example the static transactions used {update, key, type, op}, while the interactive used {update, {key, type op}}.  I changed this so now everything uses the same format which is {operation, args}, so the update is {update, {key, type, op}} and the read is {read, {key, type}}.  These are now in the type definitions in antidote.hrl

The changes from the other comments are also added.  (Although I use the emacs erlang plugin for spacing, which maybe has different rules because it didn't see a couple of the spacing problems suggested.)

riak_test/clocksi_test.erl
```diff
@@ -19,20 +19,40 @@
...
--export([confirm/0, clocksi_test1/1, clocksi_test2/1,
-         clocksi_test_read_wait/1, clocksi_test4/1, clocksi_test_read_time/1,
-         clocksi_test_certification_check/1,
-         clocksi_multiple_test_certification_check/1, spawn_read/3, clocksi_test_prepare/1, spawn_com/2]).
+-export([confirm/0,
+	 clocksi_test1/1,
+	 clocksi_test2/1,
+	 clocksi_test3/1,
+	 clocksi_test5/1,
+         clocksi_test_read_wait/1,
```
cmeiklejohn: What's with all of this wacky indenting?

riak_test/log_handoff_test.erl
```diff
@@ -40,6 +43,11 @@ confirm() ->
     [RootNode | TestNodes] = rt:deploy_nodes(Versions,[logging]),
     rt:wait_for_service(RootNode, logging),
 
+
+    lager:info("Waiting until vnodes are started up"),
```
cmeiklejohn: What happens if you don't wait?

riak_test/opcounter_test.erl
```diff
@@ -21,14 +21,24 @@
 
 -export([confirm/0, clocksi_test1/1, clocksi_test2/1,
          clocksi_test_read_wait/1, clocksi_test4/1, clocksi_test_read_time/1,
-         clocksi_test_certification_check/1,
-         clocksi_multiple_test_certification_check/1, spawn_read/3]).
+	 clocksi_test_certification_check/1,
```
cmeiklejohn: Clean this formatting up.

src/antidote.erl
```diff
@@ -110,30 +102,50 @@ clocksi_istart_tx() ->
...
     clocksi_execute_tx(ClientClock, Operations).
 
--spec clocksi_bulk_update(Operations :: [any()]) -> term().
+-spec clocksi_bulk_update([client_op()]) -> {ok, {txid(), [snapshot()], commit_time()}} | {error, term()}.
 clocksi_bulk_update(Operations) ->
     clocksi_execute_tx(Operations).
 
 -spec clocksi_read(ClientClock :: snapshot_time(),
-                   Key :: key(), Type:: type()) -> term().
+                   Key :: key(), Type:: type()) -> {ok, {txid(), [snapshot()], commit_time()}} | {error, term()}.
```
cmeiklejohn: Maybe create a type for this?  You use it all over.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -202,67 +301,118 @@ receive_prepared(abort, S0) ->
...
+			    DcId = ?DC_UTIL:get_my_dc_id(),
+			    CausalClock = ?VECTORCLOCK:set_clock_of_dc(
+					    DcId, CommitTime, Transaction#transaction.vec_snapshot_time),
+			    {ok, {TxId, CausalClock}};
+			aborted->
+			    {aborted, TxId};
+			Reason->
+			    {TxId, Reason}
+		    end,
+	    _Res = gen_fsm:reply(From,Reply);
```
cmeiklejohn: No assertion?

src/clocksi_readitem_fsm.erl
```diff
@@ -28,110 +28,238 @@
...
+handle_call({perform_read, Key, Type, Transaction},Coordinator,
+	    SD0=#state{ops_cache=OpsCache,snapshot_cache=SnapshotCache,prepared_cache=PreparedCache,self=Self}) ->
+    perform_read_internal(Coordinator,Key,Type,Transaction,OpsCache,SnapshotCache,PreparedCache,Self),
+    {noreply,SD0};
+
+handle_call({go_down},_Sender,SD0) ->
+    {stop,shutdown,ok,SD0}.
+
+handle_cast({perform_read_cast, Coordinator, Key, Type, Transaction},
+	    SD0=#state{ops_cache=OpsCache,snapshot_cache=SnapshotCache,prepared_cache=PreparedCache,self=Self}) ->
```
cmeiklejohn: Fix this formatting; this is impossible to read.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -108,92 +119,180 @@ stop(Pid) -> gen_fsm:sync_send_all_state_event(Pid,stop).
...
-            SnapshotTime = Transaction#transaction.snapshot_time,
-            gen_fsm:reply(From, {ok, SnapshotTime}),
-            {next_state, committing, SD0#state{state=committing, commit_time=SnapshotTime}};
+prepare(timeout, SD0=#state{
+                        transaction = Transaction,
+                        updated_partitions=Updated_partitions, from=From}) ->
+    case length(Updated_partitions) of
+        0->
+            Snapshot_time=Transaction#transaction.snapshot_time,
+            _Res = gen_fsm:reply(From, {ok, Snapshot_time}),
```
cmeiklejohn: I'm a bit unclear here why you choose to ignore with `_Res` when the function only has one possible return value... reasons?

cmeiklejohn:
This test is not passing for me:

```
14:55:03.564 [info] opbcounter_test prereqs: []
14:55:03.564 [notice] Running Test opbcounter_test
14:55:04.577 [info] Test Runner `uname -a` : Darwin Christophers-MacBook-Pro.local 14.4.0 Darwin Kernel Version 14.4.0: Thu May 28 11:35:04 PDT 2015; root:xnu-2782.30.5~1/RELEASE_X86_64 x86_64

14:55:04.577 [info] Riak path: "/Users/cmeiklejohn/rt/antidote"
14:55:04.638 [info] Running: /Users/cmeiklejohn/rt/antidote/current/dev/dev2/bin/antidote start
14:55:04.638 [info] Running: /Users/cmeiklejohn/rt/antidote/current/dev/dev1/bin/antidote start
14:55:04.638 [info] Running: /Users/cmeiklejohn/rt/antidote/current/dev/dev3/bin/antidote start
14:55:04.953 [error] Error in process <0.141.0> on node 'riak_test@127.0.0.1' with exit value: {badarg,[{erlang,register,[cover_server,<0.141.0>],[]},{cover,init_main,1,[{file,"cover.erl"},{line,585}]}]}


14:55:04.953 [error] Error in process <0.142.0> on node 'riak_test@127.0.0.1' with exit value: {badarg,[{erlang,register,[cover_server,<0.142.0>],[]},{cover,init_main,1,[{file,"cover.erl"},{line,585}]}]}


14:55:05.016 [info] Wait until 'dev2@127.0.0.1' is pingable
14:55:05.016 [info] Wait until 'dev3@127.0.0.1' is pingable
14:55:05.019 [info] Wait until 'dev1@127.0.0.1' is pingable
14:55:06.786 [info] Wait until 'dev1@127.0.0.1' is pingable
14:55:06.787 [info] Wait until 'dev2@127.0.0.1' is pingable
14:55:06.787 [info] Wait until 'dev3@127.0.0.1' is pingable
14:55:06.788 [info] Wait until riak_core_ring_manager is up on 'dev1@127.0.0.1'
14:55:06.789 [info] Wait until riak_core_ring_manager is up on 'dev2@127.0.0.1'
14:55:06.790 [info] Wait until riak_core_ring_manager is up on 'dev3@127.0.0.1'
14:55:06.790 [info] Check 'dev1@127.0.0.1' is a singleton
14:55:06.959 [info] Check 'dev2@127.0.0.1' is a singleton
14:55:06.959 [info] Check 'dev3@127.0.0.1' is a singleton
14:55:06.959 [info] Deployed nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:06.987 [info] [join] 'dev2@127.0.0.1' to ('dev1@127.0.0.1'): ok
14:55:06.997 [info] [join] 'dev3@127.0.0.1' to ('dev1@127.0.0.1'): ok
14:55:07.507 [info] planning and commiting cluster join
14:55:07.520 [info] plan: done
14:55:08.041 [info] Wait until nodes are ready : ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:08.043 [info] Wait until nodes agree about ownership ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:27.165 [info] Wait until no pending changes on ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:46.267 [info] Cluster built: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:46.267 [info] Nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:55:46.267 [info] new_bcounter_test started
14:55:46.651 [info] increment_test started
14:55:46.841 [info] decrement_test started
14:55:46.934 [info] transfer_test started
14:55:47.052 [info] conditional_write_test started
```

src/clocksi_interactive_tx_coord_fsm.erl
tcrain: I put this here because in the erlang docs gen_fsm:reply returns true
http://www.erlang.org/doc/man/gen_fsm.html#reply-2

But in the version of erlang I am using (Erlang R16B02 (erts-5.10.3) [source] [64-bit]) it returns the following:
Res: {#Ref<0.0.0.11727>,ok}

So maybe it is best to do a case statement to match either of these?

riak_test/log_handoff_test.erl
tcrain: If the vnodes are not started then the test will crash.  This is because the reads happen from the shared ets table that is owned by the vnode, but can be accessed by other concurrent reader threads.

An alternative could be to add the check within the read operation instead of during startup.  This would probably be better for safety.

I guess I have it the way it is currently because as far as I understand consistency isn't insured during handoff, so I am assuming when we look at how we will ensure consistency during handoff this issue will be solved as well since startup is sort of a sub issue of handoff.

riak_test/clocksi_test.erl
tcrain: It's probably from people using different editors.  I miss these because they look fine in the raw file
https://raw.githubusercontent.com/SyncFree/antidote/materializer_multidc_eval_strong/riak_test/clocksi_test.erl

I think when you press the "fix-indenting" button in emacs it just fills up the missing spaces to make it even with the other lines.  So with people using different editors and when some spaces are already there you get a mix of tab and space characters.  I fix them when I notice them.

tcrain:
Just to check, does riak test still say the test is passing, but you are getting those errors in cover.erl?

On my machine I don't get those errors, maybe we are using different dependencies? Is there a command to check that?

13:55:40.169 [info] opbcounter_test prereqs: []
13:55:40.169 [notice] Running Test opbcounter_test
13:55:41.189 [info] Test Runner `uname -a` : Darwin 14.3.0 Darwin Kernel Version 14.3.0: Mon Mar 23 11:59:05 PDT 2015; root:xnu-2782.20.48~5/RELEASE_X86_64 x86_64

13:55:41.189 [info] Riak path: "/Users/tcrain/rt/antidote"
13:55:41.296 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev1/bin/antidote start
13:55:41.296 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev2/bin/antidote start
13:55:41.296 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev3/bin/antidote start
13:55:43.488 [info] Wait until 'dev1@127.0.0.1' is pingable
13:55:43.489 [info] Wait until 'dev2@127.0.0.1' is pingable
13:55:43.489 [info] Wait until 'dev3@127.0.0.1' is pingable
13:55:46.122 [info] Wait until 'dev1@127.0.0.1' is pingable
13:55:46.122 [info] Wait until 'dev2@127.0.0.1' is pingable
13:55:46.123 [info] Wait until 'dev3@127.0.0.1' is pingable
13:55:46.123 [info] Wait until riak_core_ring_manager is up on 'dev1@127.0.0.1'
13:55:47.128 [info] Wait until riak_core_ring_manager is up on 'dev2@127.0.0.1'
13:55:47.129 [info] Wait until riak_core_ring_manager is up on 'dev3@127.0.0.1'
13:55:47.130 [info] Check 'dev1@127.0.0.1' is a singleton
13:55:47.224 [info] Check 'dev2@127.0.0.1' is a singleton
13:55:47.224 [info] Check 'dev3@127.0.0.1' is a singleton
13:55:47.225 [info] Deployed nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
13:55:47.423 [info] [join] 'dev2@127.0.0.1' to ('dev1@127.0.0.1'): ok
13:55:48.084 [info] [join] 'dev3@127.0.0.1' to ('dev1@127.0.0.1'): ok
13:55:48.585 [info] planning and commiting cluster join
13:55:51.568 [info] Nodes not ready after initial plan/commit, retrying
13:55:52.070 [info] planning and commiting cluster join
13:55:52.490 [info] Wait until nodes are ready : ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
13:55:53.498 [info] Wait until nodes agree about ownership ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
13:58:55.115 [info] Wait until no pending changes on ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:00:02.411 [info] Cluster built: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:00:02.411 [info] Nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:00:02.411 [info] new_bcounter_test started
14:00:02.762 [info] increment_test started
14:00:02.785 [info] decrement_test started
14:00:02.790 [info] transfer_test started
14:00:02.794 [info] conditional_write_test started
14:00:02.841 [info] Stopping riak on 'dev1@127.0.0.1'
14:00:12.846 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev1/bin/antidote stop
14:00:24.683 [info] Wait until 'dev1@127.0.0.1' is not pingable
14:00:24.684 [info] Stopping riak on 'dev2@127.0.0.1'
14:00:34.688 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev2/bin/antidote stop
14:00:46.235 [info] Wait until 'dev2@127.0.0.1' is not pingable
14:00:46.237 [info] Stopping riak on 'dev3@127.0.0.1'
14:00:56.241 [info] Running: /Users/tcrain/rt/antidote/current/dev/dev3/bin/antidote stop
14:01:06.760 [info] Wait until 'dev3@127.0.0.1' is not pingable
14:01:06.762 [info] Removing directory /Users/tcrain/rt/antidote/current/dev/dev1/data/
14:01:06.786 [info] Removing directory /Users/tcrain/rt/antidote/current/dev/dev2/data/
14:01:06.798 [info] Removing directory /Users/tcrain/rt/antidote/current/dev/dev3/data/
14:01:06.809 [notice] opbcounter_test Test Run Complete
14:01:06.810 [info] Stopping cover
14:01:06.810 [info] Multiple tests run or no failure
14:01:06.810 [info] Trying to obtain node shutdown_time via RPC...
14:01:06.811 [info] Using node shutdown_time of 20000
14:01:06.813 [info] RPC to node 'dev3@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev3/bin/riak stop
14:01:06.813 [info] RPC to node 'dev4@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev4/bin/riak stop
14:01:06.813 [info] RPC to node 'dev2@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev2/bin/riak stop
14:01:06.813 [info] RPC to node 'dev5@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev5/bin/riak stop
14:01:06.813 [info] RPC to node 'dev6@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev6/bin/riak stop
14:01:06.813 [info] RPC to node 'dev1@127.0.0.1' returned {badrpc,nodedown}, will try stop anyway... /Users/tcrain/rt/antidote/current/dev/dev1/bin/riak stop
14:01:06.820 [info] Stopped node 'dev3@127.0.0.1', stop status: wasn't running.
14:01:06.821 [info] Stopped node 'dev4@127.0.0.1', stop status: wasn't running.
14:01:06.822 [info] Stopped node 'dev5@127.0.0.1', stop status: wasn't running.
14:01:06.822 [info] Stopped node 'dev2@127.0.0.1', stop status: wasn't running.
14:01:06.824 [info] Stopped node 'dev6@127.0.0.1', stop status: wasn't running.
14:01:06.825 [info] Stopped node 'dev1@127.0.0.1', stop status: wasn't running.

Test Results:
opbcounter_test-error: pass

bieniusa:
For me the test does not pass, too.
Manuel mentioned yesterday that the reason is likely a (disabled) certification check.

Annette Bieniusa
TU Kaiserslautern
FB Informatik
PO Box 3049
67653 Kaiserslautern

bieniusa@cs.uni-kl.de <mailto:bieniusa@cs.uni-kl.de>

Office:Building 34, Room 416
Phone:+49 (0) 631 205 3958
Fax:+49 (0) 631 205 3420

tcrain:
Can you show the error message please?  The certification checks should be enabled, unless I broke something.

bieniusa:
In my case, it simply hangs:

14:29:16.178 [info] Riak path: "/Users/annette/rt/antidote"
14:29:16.229 [info] Running: /Users/annette/rt/antidote/current/dev/dev1/bin/antidote start
14:29:16.229 [info] Running: /Users/annette/rt/antidote/current/dev/dev2/bin/antidote start
14:29:16.229 [info] Running: /Users/annette/rt/antidote/current/dev/dev3/bin/antidote start
14:29:16.505 [info] Wait until 'dev3@127.0.0.1' is pingable
14:29:16.505 [info] Wait until 'dev2@127.0.0.1' is pingable
14:29:16.505 [info] Wait until 'dev1@127.0.0.1' is pingable
14:29:17.937 [info] Wait until 'dev1@127.0.0.1' is pingable
14:29:17.937 [info] Wait until 'dev2@127.0.0.1' is pingable
14:29:17.937 [info] Wait until 'dev3@127.0.0.1' is pingable
14:29:17.937 [info] Wait until riak_core_ring_manager is up on 'dev1@127.0.0.1'
14:29:17.938 [info] Wait until riak_core_ring_manager is up on 'dev2@127.0.0.1'
14:29:17.939 [info] Wait until riak_core_ring_manager is up on 'dev3@127.0.0.1'
14:29:17.939 [info] Check 'dev1@127.0.0.1' is a singleton
14:29:18.059 [info] Check 'dev2@127.0.0.1' is a singleton
14:29:18.060 [info] Check 'dev3@127.0.0.1' is a singleton
14:29:18.061 [info] Deployed nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:18.078 [info] [join] 'dev2@127.0.0.1' to ('dev1@127.0.0.1'): ok
14:29:18.100 [info] [join] 'dev3@127.0.0.1' to ('dev1@127.0.0.1'): ok
14:29:18.601 [info] planning and commiting cluster join
14:29:19.135 [info] Wait until nodes are ready : ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:19.136 [info] Wait until nodes agree about ownership ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:38.186 [info] Wait until no pending changes on ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:58.266 [info] Cluster built: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:58.266 [info] Nodes: ['dev1@127.0.0.1','dev2@127.0.0.1','dev3@127.0.0.1']
14:29:58.266 [info] new_bcounter_test started
14:29:58.938 [info] increment_test started
14:29:59.303 [info] decrement_test started
14:29:59.422 [info] transfer_test started
14:29:59.541 [info] conditional_write_test started

tcrain:
Ok, can you check the rt/antidote/current/dev/dev1,2,3/log/console.log logs for errors?  The test works for me on two different machines so maybe it is a concurrency issue, I will check if I see something in the code.

tcrain:
Please review again.  I have made some bug fixes, and performance fixes:

- More efficient checking and waiting when there are concurrent transactions being prepared
- Fixed a bug in prepare that was causing the bounded counter test to hang
- Added option to use synchronous logging in antidote.hrl.  This is just a couple lines that will ensure the full transaction log is written to disk when the commit record is stored.
- There were some supervisors that were being supervised as workers even though they were supervisors, was there a reason for this? I changed this so that they are supervised as supervisors: https://github.com/SyncFree/antidote/commit/df26e7635a02bae5d35788bf2e423b023324e1ba
- Made static tx coords their own fsm (see https://github.com/SyncFree/antidote/pull/110 it is the same idea as this pull request, but updated for this branch)
- Made it so it is safe to read from ets tables even if vnodes have not been started up (the problem solved here is that vnodes are not started up right away, and since ets tables are shared to allow parallel reads a table might have been read before is was initialised, but now a check is done first and if the vnode is not started yet, the request is forwarded directly to vnode master, forcing it to start).  I tried to solve this for corner cases as well, but keep an eye out in case you think I missed some.

tcrain:
PS I added one last optimisation to this branch, a fast path for a single update operation.  This will allow us to check the overhead of transactions vs single ops.

PPS I just realised I forgot to run dialyzer so hold off on the review for a day please.

tcrain:
Please review when you can, dialyzer and all tests should be passing

tcrain:
Heres a couple performance numbers for single DC doing small scale single operation benchmarks.
(These are with the SI validation check and disk logging disabled since there are some issues with those that will be resolved in future updates).
The nodes used are each: (2 CPUs Intel Xeon E5-2630 v3, 8 cores/CPU, 125GB RAM, 558GB HDD, 10Gbps ethernet) [all nodes are in a single grid5000 site, cross site and multi-dc benchmarks are also possible]

y-axis is operations per second.
x-axis is varying the percentage of read/update operations

Graph 1key-1nodecluster.png is a single node cluster using only a single key.  This means that every concurrent update is "conflicting but mergeable".  There are 80 clients run on a separate node.  The idea here is to test a single CRDT/single key scalability.

Graph 2000key-3nodecluster.png is a 3 node cluster using 2000 keys.  There are 560 clients across 7 separate nodes. (I also added the read and update latencies for this benchmark).

Each key contains an op-based or-set.  Each or-set has approximately 10 elements (plus/minus several elements depending on concurrency), with each element being of size 100 bytes.

A read returns a single or-set, or approx 1000 bytes.

An update first performs a read, then performs an add or removal (so it contains 2 round trips to the server and writes are more expensive so latency is higher).

One thing to note is that a single node cluster by itself will be faster (per node) than a multi-node cluster because the key being searched for will always be on that node (nodes are chosen at random by the client when it sends the request).  For example with a 3 node cluster there is a 66% chance of the request needing to be forwarded to another node (meaning another round trip through the network).

These benchmarks and graphs can all be generated in a just a couple commands once you get a grid5000 environment set up using the scripts here https://github.com/SyncFree/basho_bench/tree/grid5000 , instructions here https://github.com/SyncFree/basho_bench/blob/grid5000/GRID5000_EXAMPLE
So I encourage you to play with and break stuff.

1key-1nodecluster

2000keys-3nodecluster

read_latencies

write_latencies

marc-shapiro:
How does the performance compare with the previous version, without optimisations?

					Marc

tcrain:
I couldn't get ahold of the fast machines, because they are under maintenance, but I reran the benchmarks on these slower nodes: (2 CPUs Intel Xeon E5520, 4 cores/CPU, 31GB RAM, 557GB HDD)
Also I couldn't get the 3 node cluster to commit in a reasonable amount of time in the old branch so I just ran it using single node clusters.  All are done with SI certification-check disabled and with no disk-logging.

One thing to note is that the main goal of these changes were for increasing concurrency so that is where the most performance improvement will be (i.e. single node, single key updates have the least improvement), so there are still lots of other things that could be done in other parts.

Another thing to note is that Erlang doesn't have the best data-structure support for certain parts of the multi-versioning, concurrent updating, shared data-base kind of thing.  Maybe as a future work we can look into making our own shared-memory data structures using compare and swap.  Multi-versioning with vector clocks and CRDTs is very expensive so making custom shared non-blocking data-structures should be interesting research and improve performance several x.

Here is the old performance for 80 clients, 2000 keys.
summary_overall

read_latencies

write_latencies

(Note that the low write latency for 99.99(.01) is an artifact of the way the data is collected so it should be up in line with the 99(1) value)

Here is with the changes:
summary_overall

read_latencies

write_latencies


Here is the single key configuration with 80 clients, old version:
summary_overall

read_latencies

write_latencies

(Note that the low write latency for 99.99(.01) is an artifact of the way the data is collected so it should be up in line with the 99(1) value)


And with the changes:
summary_overall

read_latencies

write_latencies

aletomsic:
great!

marc-shapiro:
So if I just look at the first three pairs of graphs (2000 keys), I conclude:
	1) Throughput improves from 3x (on the right) to 13x (on the left)
	2) Common case latency improves by 15x, worst-case is more or less the same as before
	3) Common case latency improves by 5x, worst-case by 2x

Is this correct?

What is the difference between graphs 2 and 3?

				Marc

tcrain:
Approx 8x on the left.

2 is read latencies, 3 write latencies

marc-shapiro:
Thx

		Marc

Issue #140, opened 2015-09-12
clocksi_downstream: remove case per op_based crdt


aletomsic:
Now the code requiring a call to downstream generaton per op_based crdt, and looks like this.
I suggest having one per riak_dt, and one for all op_based.

DownstreamOp = case Type of
                                   crdt_bcounter ->
                                       case Type:generate_downstream(Op, Actor, Snapshot) of
                                           {ok, OpParam} -> {update, OpParam};
                                           {error, Error} -> {error, Error}
                                       end;
                                   crdt_orset ->
                                       {ok, OpParam} = Type:generate_downstream(Op, Actor, Snapshot),
                                       {update, OpParam};
                                   crdt_pncounter ->
                                       {ok, OpParam} = Type:generate_downstream(Op, Actor, Snapshot),
                                       {update, OpParam};
                                   _ ->
                                       {ok, NewState} = Type:update(Op, Actor, Snapshot),
                                       {merge, NewState}
                               end,

aletomsic:
addressed in pull request #142.

check (only) case in clocksi_downstream

Issue #84, opened 2014-11-26
Type Check


aletomsic:
The way I see it, there are two possible directions to take from here.

The first one, which I think is the best in terms of performance and elegance, would be to perform the  type check as soon as a transaction starts. So, what this means is that the type check is the first task a tx coordinator should perform. So, the tx coord would call materialzer:type_check(Operations). The materialiser should check if for each op, the parameters sent are valid. The current implementation presents two problems for performing a complete check:
it does not embed the type in the key, so at this stage we can not check that the operations also respect that the type is the same as the one stored for that key (if it exists).
it does not include the type in update operations.

The second one (and easiest from the point the implementation is now) would be to perform the type check when generating the downstream operation. This presents the need of reading the key first in order to know its type and then see if the operation is supported and the parameters sent by the update are valid. Checking at this point would be expensive, as maybe many operations could have already generated its downstream. All the process of starting the coordinator, sending correct updates to partitions and then aborting the transaction will be done because of a wrong typed or bad parametrised operation.

I could start by doing the second approach (which could be pretty easy) and then discuss about implementing the first one. I await your comments.

aletomsic:
Proposed solution for the new interface (see Deepthi’s thread “Andidote application/materialiser interface”):


If I understand correctly, the new interface will:

1) create buckets containing objects of the same type. How will this be implemented?
2) bind an object on its creation to a certain type. 
3) read and update operations will not take as a parameter the object type, as it will be embedded on the object's bucket.

Solution:
Whenever an object is handled inappropriately, a transaction will abort.
Inappropriate handling includes (please contribute to completing this list if I am missing anything):

* binding a key to a type that does not exist.
* using an undefined method for a given type, e.g. trying to increment a set.
* using a defined method's type with incorrect parameters.
* issuing an operation on a not bound key.

Remark: binding an already bound key simply generates a new key in a different bucket.


Implementation:


binding a key to a type that does not exist.

issue a type:new() operation at the coordinator. In case of success, continue; otherwise, abort the txn.

using an undefined method for a given type, e.g. trying to increment a set.
using a defined method's type with incorrect parameters.

for these cases, issue a type:operation(params) at the coordinator. Abort if fails.

issuing an operation on a not bound key.

this is the only tricky check, as it requires checking that the object exists, which means contacting the partition that stores the object. 
my suggestion is to perform this check optimistically, at downstream generation. It could be too late, but this way we would not incur in significant overhead.
[...]

PR #142, opened 2015-09-13
Type check

aletomsic:
This pull request tacles issues: 
- clocksi_downstream: remove case per op_based crdt #140
- Type Check #84 
The type check should be revised once the new interface is in place, as it is interface dependent.

### DIFF
--- a/src/crdt_bcounter.erl
+++ b/src/crdt_bcounter.erl
@@ -16,13 +16,13 @@
 
 %% API
 -export([new/0,
-         localPermissions/2,
-         permissions/1,
-         value/1,
-         generate_downstream/3,
-         update/2,
-         to_binary/1,
-         from_binary/1]).
+    localPermissions/2,
+    permissions/1,
+    value/1,
+    generate_downstream/3,
+    update/2,
+    to_binary/1,
+    from_binary/1, is_operation/1]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -154,6 +154,28 @@ to_binary(C) -> term_to_binary(C).
 -spec from_binary(binary()) -> bcounter().
 from_binary(<<B/binary>>) -> binary_to_term(B).
 
+%% @doc The following operation verifies
+%%      that Operation is supported by this particular CRDT.
+-spec is_operation(term()) -> boolean().
+is_operation(Operation) ->
+    case Operation of
+        {decrement, Number} ->
+            is_integer(Number) and (Number >= 0);
+        {increment, Number} ->
+            is_integer(Number) and (Number >= 0);
+        {transfer, Number, _} ->
+            is_integer(Number) and (Number >= 0);
+        {{decrement, Number}, _} ->
+            is_integer(Number) and (Number >= 0);
+        {{increment, Number}, _} ->
+            is_integer(Number) and (Number >= 0);
+        {{transfer, Number, _}, _} ->
+            is_integer(Number) and (Number >= 0);
+        _ ->
+            false
+    end.
+
+
 %% ===================================================================
 %% EUnit tests
 %% ===================================================================
@@ -256,4 +278,14 @@ binary_test() ->
     %% Assert marshaling and unmarshaling holds the same `bcounter()'.
     B = to_binary(Counter3),
     ?assertEqual(Counter3, from_binary(B)).
+
+is_operation_test() ->
+    ?assertEqual(true, is_operation({transfer, 2, r2})),
+    ?assertEqual(true, is_operation({increment, 50})),
+    ?assertEqual(false, is_operation(increment)),
+    ?assertEqual(true, is_operation({decrement, 50})),
+    ?assertEqual(false, is_operation(decrement)),
+    ?assertEqual(false, is_operation({anything, [1,2,3]})).
+
+
 -endif.
--- a/src/crdt_bcounter.erl
+++ b/src/crdt_bcounter.erl
@@ -22,7 +22,8 @@
     generate_downstream/3,
     update/2,
     to_binary/1,
-    from_binary/1, is_operation/1]).
+    from_binary/1, 
+    is_operation/1]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -160,17 +161,17 @@ from_binary(<<B/binary>>) -> binary_to_term(B).
 is_operation(Operation) ->
     case Operation of
         {decrement, Number} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         {increment, Number} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         {transfer, Number, _} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         {{decrement, Number}, _} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         {{increment, Number}, _} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         {{transfer, Number, _}, _} ->
-            is_integer(Number) and (Number >= 0);
+            is_integer(Number) andalso (Number >= 0);
         _ ->
             false
     end.
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ clean:
 	$(REBAR) clean
 	rm -rf riak_test/ebin
 
-distclean: clean devclean relclean
+distclean: clean devclean relclean cleanplt
 	$(REBAR) delete-deps
 
 rel: all
--- a/README.md
+++ b/README.md
@@ -54,13 +54,7 @@ list), then you should be able to start a node of `antidote`.
     15:55:05:antidote $ rel/antidote/bin/antidote console
     (elided)
     Eshell V5.10.3  (abort with ^G)
-    (antidote@127.0.0.1)1> antidote:ping().
-    {pong,1118962191081472546749696200048404186924073353216}
-    (antidote@127.0.0.1)3>
-
-What you should see is a `pong` response, followed by a big number. The
-number is the partition that responded to the `ping` request. Try it a
-few more times, different partitions will respond.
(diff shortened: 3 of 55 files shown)


tcrain:
Just a suggestion: it looks like you ran clocksi_vnode.erl and clocksi_interactive_tx_coord_fsm.erl through something that fixes spacing.  Could you maybe run that directly on master and then make this pull request from there (or do a separate spacing pull request for the spacing fixes).  It should take 2 minutes and then the diffs will be readable.  (Probably a lot of the spacing errors are mine, so maybe you can also show me what you use to fix them for the future also)

Also maybe give a short description of what the type check does, because the discussion in https://github.com/SyncFree/antidote/issues/84 has a lot of different paths of discussion.

aletomsic:
If everyone agrees, I can try to run the same thing that fixes spacing in every branch, but that would need everyone to merge with the open branches that exist. 

Regarding the changes I've made, I'm now about to comment it on the issue's thread.

bieniusa:
I don’t agree. 
Every developer needs to be able to maintain a branch with proper indentation.

Please just fix your own branch (and maybe check the master branch).

	Annette

aletomsic:
> On Sep 14, 2015, at 13:01, Annette Bieniusa <notifications@github.com> wrote:
> 
> I don’t agree. 
> Every developer needs to be able to maintain a branch with proper indentation.
> 
> Please just fix your own branch (and maybe check the master branch).

Sorry, I meant muster, and not every brunch.
Sure.

tcrain:
Maybe someone can suggest a tool for everyone to use for spacing since different editors use different conventions?  Something that we can run on the master branch first, then run on our branches before we do a pull request.

aletomsic:
I have found an issue with opbcounter_test, closing for now, will reopen when fixed.

aletomsic:
I've checked the map and fold alternatives, but they iterate over all the elements of the list, unless there is an exception.
In our case, what we want to try is to stop the iteration once we find a wrong operation. 
please let me know what you think.

Regarding the alternative way of type checking, I can issue it in a future pull request, once I investigate that, so that at least we have a form of type checking in place.


I can't find in riak_dt a method that checks that an operation is correctly typed. 
Someone mentioned that was already in place in riak_dt, Valter? Could you point me to it?

cmeiklejohn:
In Riak itself, this isn't enforced in Riak DT, it's enforced at the integration point between the user API and the data type.

marc-shapiro:
But how do they do it?

		Marc

cmeiklejohn:
There's a series of validations here that parse an incoming operation in JSON to a valid operation for the given data type.  

https://github.com/basho/riak_kv/blob/develop/src/riak_kv_crdt_json.erl

For instance, the following for registers:

https://github.com/basho/riak_kv/blob/develop/src/riak_kv_crdt_json.erl#L175

aletomsic:
chris, do you think that the way I'm type checking introduces significant overhead?

basically, what I do is a type:update(OpParams, Type:new()) in the case of a state based crdt, 
and a type:generate_downstream(OpParams, Type:new()) in case of an op_based one.

Otherwise, which approach would you suggest?

cmeiklejohn:
Can you link me the actual code?

aletomsic:
https://github.com/SyncFree/antidote/blob/e54a21de26a61587f36ea6e553485af00f54ef3f/src/materializer.erl#L63

cmeiklejohn:
It's most likely faster to just store a list of valid mutations along with the data type and check that the operation you've received is in the list, rather than attempting to apply the update and catching the exception.  Maybe you can extend Riak DT with this information...

bieniusa:
Which is exactly what we discussed and decided in today's meeting.

Annette

Von meinem iPhone gesendet

cmeiklejohn:
I'd probably extend Riak DT with this so all users of the library can take advantage of it.

aletomsic:
I’m afraid that coming up with a full list of possible crdt operations will take time.
I would prefer to use that time to write the paper and work on implementing the NMSI protocol i’m working on.
As this current solution works, I’ll benchmark this solution when compared to master and see if it introduces overhead.
Then, I’ll create a check for a couple of crdts and compare them, then we can decide.
Ok?

cmeiklejohn:
Why will it take time?  There's already a list in Riak DT of the supported operations that is used for type analysis via Dialyzer.  Can't you just copy this and make a function that returns the types?  Isn't this only a couple lines of code for each data type?

https://github.com/basho/riak_dt/blob/develop/src/riak_dt_orset.erl#L52

marc-shapiro:
This will work, and is probably a lot faster than Alejandro's alternative.  But it still adds overhead on top of every single object invocation.

Wouldn’t it be possible to check, just once, the first time the object is used (or at least when it is brought into the materialiser cache), that it is bound with the correct implementation?  

								Marc

bieniusa:
Am 15.09.2015 um 19:49 schrieb marc-shapiro <notifications@github.com>:
> 
> This will work, and is probably a lot faster than Alejandro's alternative. But it still adds overhead on top of every single object invocation.
> 
> Wouldn’t it be possible to check, just once, the first time the object is used (or at least when it is brought into the materialiser cache), that it is bound with the correct implementation? 
> 
> Marc

I think the way we envisioned the interface it is unavoidable to check every single operation for type compliance (dynamically).
A programmer might very well correctly insert an element into a CRDT set as the first operation in a transaction, and then try to increment the set by 3.

I think Ale can make this without too much effort configurable.

	Annette

bieniusa:
> Am 15.09.2015 um 19:27 schrieb Alejandro Zlatko Tomsic <notifications@github.com>:
> 
> I’m afraid that coming up with a full list of possible crdt operations will take time.
> I would prefer to use that time to write the paper and work on implementing the NMSI protocol i’m working on.
> As this current solution works, I’ll benchmark this solution when compared to master and see if it introduces overhead.
> Then, I’ll create a check for a couple of crdts and compare them, then we can decide.
> Ok?


No, this is not ok.
We have discussed in today’s meeting extensively why this solution is actually not working as we want.
I don’t see any reason why benchmarking it would be useful - this is indeed a waste of time.

Please just stick to the plan, this will provide a good solution in little time.

	Annette

aletomsic:
> On Sep 16, 2015, at 00:52, Annette Bieniusa <notifications@github.com> wrote:
> 
> 
> > Am 15.09.2015 um 19:27 schrieb Alejandro Zlatko Tomsic <notifications@github.com>:
> > 
> > I’m afraid that coming up with a full list of possible crdt operations will take time.
> > I would prefer to use that time to write the paper and work on implementing the NMSI protocol i’m working on.
> > As this current solution works, I’ll benchmark this solution when compared to master and see if it introduces overhead.
> > Then, I’ll create a check for a couple of crdts and compare them, then we can decide.
> > Ok?
> 
> 
> No, this is not ok.
> We have discussed in today’s meeting extensively why this solution is actually not working as we want.
> I don’t see any reason why benchmarking it would be useful - this is indeed a waste of time.
> 
> Please just stick to the plan, this will provide a good solution in little time.

I see now that doing it as discussed yesterday is simple. I’ll take care of it before next meeting.
It is, anyway, frustrating that I’ve sent (twice) emails explaining the approach I was gonna take to implement it. I even sent one saying “starting now with this approach”.  I received no feedback.
Only when I asked a pull request, there was (negative) reaction. 
It took me 2 days (that are now lost) to implement the “wrong” type checking as I run into cases hard to debug and running the test suit takes time.
It would be nice for future times to get feedback when requested, before going in the wrong direction.

aletomsic:
Chris, Annette,

Just to see that I am getting it right. 
I have created the following code in riak_dt:

-define(RIAK_DT_TYPES,
    [riak_dt_disable_flag, riak_dt_emcntr, riak_dt_enable_flag,
    riak_dt_gcounter, riak_dt_gset, riak_dt_lwwreg, riak_dt_map,
    riak_dt_od_flag, riak_dt_oe_flag, riak_dt_orset, riak_dt_orswot,
    riak_dt_pncounter, riak_dt_vclock]).

is_riak_dt(Atom) ->
    lists:member(Atom, ?RIAK_DT_TYPES).



Then, for instance, for a gcounter, i have the following:

is_operation(Operation) ->
    case Operation of
        increment ->
                    true;
        decrement ->
                    true;
        {increment, Number} ->
                    is_integer(Number);
        {decrement, Number} ->
                    is_integer(Number);
        _ ->
            false
    end.

Is this what you meant?

bieniusa:
Yes, this fits my assumptions.

Two remarks:
- I probably would call the function type_check instead of is_operation.
- More important than the riak_dt types would be our own op-based CRDTs (crdt_bcounter, crdt_orset, crdt_pncounter and riak_dt_mvreg (strange naming convention…, but leave it like this for the moment) ).

	Annette

aletomsic:
Hi,

I've changed the type checking to work as discussed in the last meeting. 
Now, every CRDT, both op_based and riak_dt, has a crdt:is_operation(OpParams) method that takes care of checking that an update operation is being called with the right parameters.
Whenever an operation is issued by a transaction, this check is performed. 
For a read operation, it checks that the Type is an existing CRDT type, as defined by riak_dt or our own op_based crdt implementations. 
All tests and dialyzer checked.

IMPORTANT: to check this branch, one needs to change the rebar.config to use as riak_dt dependency the new branch I made with the type checking. The rebar.config file in this branch already includes those changes, so don't forget to repull the riak_dt deps when building. 

Question: How do I proceed with the new branch of riak_dt? Should I try to merge it with basho? Do we add it to the syncfree repo (I don't have permissions to do that)?

src/crdt_bcounter.erl
```diff
@@ -154,6 +155,28 @@ to_binary(C) -> term_to_binary(C).
 -spec from_binary(binary()) -> bcounter().
 from_binary(<<B/binary>>) -> binary_to_term(B).
 
+%% @doc The following operation verifies
+%%      that Operation is supported by this particular CRDT.
+-spec is_operation(term()) -> boolean().
+is_operation(Operation) ->
```
bieniusa: Please use andalso/2 instead of and/2 for short-circuiting here (and in the other type checks)!

aletomsic: Done in antidote, will continue in riak_dt.
Also fixed the failing test.

aletomsic: done in riak_dt.

PR #143, opened 2015-09-14
Reduce time tests

angbrav:
I have merged the changes with the latest changes in the master branch.

I have passed all the tests.

Main changes are related to the number of vnodes spawned per node. I have parametrized that in every test. Now it uses 8 by default. It can be modified by adding the following line to the riak_test config file:

%% The total number of virtual nodes to deploy in test by default.
    {num_vnodes, 8},

Riak_test spends a lot of time, creating cluster, so from now on I would say is better to put test affecting the same part of the code together. For instance, I merged append_test and append_failure_test together, so the riak_test has to only create the cluster once for both.

One more thing, I would say is better to use three nodes in the cluster (unless your test requires more) than six, because of the same reason.

Finally, I have also added a new command into the Makefile called: stage-riak-test, which assumes stagedevrel was previously made.

I think that's all, I do not know how much we have improved w.r.t. the last version of master (since Tyler already included some of these modifications in some test), but I do not think there are many more things we can do.

Manuel


riak_test/oporset_test.erl
```diff
@@ -129,31 +117,3 @@ remove_test(Nodes) ->
     Result7=rpc:call(FirstNode, antidote, read,
                     [Key, Type]),
     ?assertMatch({ok, []}, Result7).
-    
-
-concurrency_test(Nodes) ->
```
bieniusa: Why is this test no longer needed?

riak_test/opbcounter_test.erl
```diff
@@ -33,7 +37,7 @@ confirm() ->
     decrement_test(Nodes),
     transfer_test(Nodes),
     conditional_write_test(Nodes),
-    rt:clean_cluster(Nodes),
+    %rt:clean_cluster(Nodes),
```
bieniusa: Why is the call to rt:clean_cluster no longer needed? (Here and in the other test cases..)

riak_test/oporset_test.erl
angbrav: concurrency_test you mean?

That s a test that was actually testing nothing. I asked Li, who wrote it, and he told me it was created when there was no interdc module and he wrote it as a reminder, but it was actually never updated.

Manuel

riak_test/opbcounter_test.erl
angbrav: It was just in some tests. I decided to remove it since it was not affecting the tests (from what I know) and it was slowing it down.

Manuel

bieniusa: I think we still need it - at the end of every independent test case. 
Tests must be independent, and cleaning up the test fixture is essential for this.
In the long run, someone will otherwise introduce "bugs" there.

angbrav: I can added, but it will slow down everything. I think a simple way to avoid cleaning the cluster is to use different keys in the tests. (we could use the module_name + counter). Just a suggestion.

Manuel

bieniusa: This is what I said: Using different keys will break, because programmers are not careful. 
Please re-add the clean up.

   Annette

angbrav: To all test or only to the test I removed it?

bieniusa: To all (i.e. after every independent test instance).

	Annette

cmeiklejohn: > but it will slow down everything

Of course it does, it's responsible for ensuring that no data exists between tests in the same module that would cause the test to fail because of some stale state.  It's doing work, so of course it's going to slow things down.

Riak Test only ensures that state is clean *between test modules*, not *between multiple tests in the same module*.  If you have multiple tests in the same module, of which Antidote has many, this call is required to ensure *determinism*.  

You have two choices here: if the test is going to test multiple things and reconfigures the cluster between tests, this call is required to ensure that the state in the data directory has been reset -- otherwise you run the risk of executing something with the previous ring or reading data from a previous execution if you happen to access the same key.  Your other choice is to break things out into separate test modules.

But, I don't believe the call should just be removed without proper investigation.

angbrav: I was just pointing out that being careful when designing the test, we could avoid using the call most of the times. Even saving 30 seconds in total is a great saving knowing that we run the whole test suit pretty often.

cmeiklejohn: I completely agree with you.  My major point was that we need to look into each test before removing the call to determine if it will introduce any accidental non-determinism.

angbrav: Dear all,

After our discussion, I run the following test:

I modified the clocksi_test by inserting clean_cluster and build_cluster between tests (a total of 15 tests in this module)
I created a version where I carefully pick the keys to not overlap across tests in the module and thus avoid using the clean and build calls.

This are the results:

More than 20 minutes for the one with the calls
Around 50s for the other one

How should I proceed?

Best,
Manuel

tcrain:
This might more have to do with riak_core, but I have always had a problem with the tests that require clusters with several nodes to converge (for example opcounter_test uses a 3 node cluster).  In these tests sometimes the ring never converges even though no updates are happening, for example I get the following messages repeating forever:

15:42:14.277 [info] commit: plan changed
15:42:14.886 [info] planning and commiting cluster join
15:42:14.887 [info] commit: plan changed

Other times I get this message and then nothing happens:

15:54:18.775 [info] plan: ring not ready
15:54:18.880 [info] Wait until no pending changes on ['dev1@127.0.0.1']

(And I don't get any error messages in the log)
It doesn't happen every time, but sometimes it can happen over and over again and gets especially frustrating when trying to run the whole suite of tests with fingers crossed.
My computer is really slow so maybe the joining algorithm allows some sort of livelock when the scheduling is unlucky?

aletomsic:
This happens to me too.

bieniusa:
Last question before merge:
Shouldn't we also reduce the number of built nodes from 6 to 3 in the Makefile?

bieniusa:
@angbrav , I know that you are working on the configuration, but I want to progress with the other pull requests and this will speed up the reviewing.
So, I keep the branch and you open another pull request once you found time to enhance / configure the testing.

PR #144, opened 2015-09-14
fixed indentation in src/ riak_test/ and include/ folders.

aletomsic:
added beauty.


bieniusa:
Sorry, but you have to go through the changes and fix them by hand.
Your adaptations break the coding conventions in many places.

aletomsic:
Just dismiss this pull request. I suggest we agree in an auto-formatting tool, as tyler said.

cmeiklejohn:
Autoformatting tool?  

If you use the Erlang mode for VIM and/or Emacs it will indent code correctly, and you won't need to format after the fact.  The way I see it, that's the fundamental problem here -- people *still* haven't addressed problems with their editors.

aletomsic:
I just switched to IntelliJ which is really nice, it checks for error in the code on the fly, integrates very simply with git and,  with one click, you can reformat the whole code (all files). I thought it would have been nice to get it done as now all files diverge in their formatting. 

Would you point me to the coding conventions we should respect? 
I can see if I can change my configuration to respect them and then retry doing the same.

PR #147, opened 2015-09-15
Pubsub interDC, metadata handling

Eipifi:
The pull includes the rewritten interDC layer, Tyler's changes to metadata handling (no more riak_core_metadata, removed vectorclock_vnode, comments from Tyler in the following post), as well as various bug fixes. All the tests are passing.

The main changes in the interDC layer:
- ZeroMQ replaces raw TCP as the new message delivery method between DCs,
- removed inter_dc_sender and inter_dc_rcvr (single point of failure and a possible bottleneck),
- reading from the log is now done only in case of a message loss, and not in normal operation,
- pub/sub approach to interDC communication - DCs publish transactions using a best-effort broadcast provided by ZeroMQ, and other DCs subscribe to transaction streams from other DCs,
- transactions are now sent separately for each partition, balancing the load and reducing synchronization overhead.

EDIT: resubmitted with merged branch weak_meta_data_cert.


tcrain:
The idea of the meta-data changes is to replace riak-core-meta-data with something more appropriate for calculating the stable time.  The stable time is the min of all observed update timestamps from all nodes within a DC.  This needs to be calculated frequently or else stale data is observed by client transactions.

Basically the new meta-data gives only enough guarantees for this functionality.  It doesn't care about lost messages, persistence, crash and restart from a blank state, it just needs to be a lightweight collection and merging of values.  To do this, there is one fsm running on each erlang node, each vnode on this node will write its observed times to an ets table, the fsm will then merge these values and broadcast them to all other erlang nodes in this dc.  At the receiver end, this data is merged and then can be read by the vnodes on the node from an ets table.  Since for safety all nodes need to eventually participate to increase the time, if a read is done before all nodes have participated you get the initial value.

The merge functions are provided as input to the fsm for collection and sending, so a different function could be used (to take the max for example).  I added this because I will use (slightly) different functions for the partial replication.

The merge functions are in stable_time_functions.erl and the sending/collecting stuff is in meta_data_sender.erl and meta_data_manager.erl 

Thing to add:
Right now on every update it checks to make sure the configuration of nodes/vndoes hasn't changed for the check to be sure all nodes have participated, but this should only be done when there was a change, this will be changed.

Send and merge in a tree structure to save work.

Performance:

Here is a 2 dc benchmark with 2 nodes using 2000 keys as counters.  Throughput x read/update ratio.  With riak-core-meta-data and with the new meta-data.  Even though the meta-data calculation is done in the background, using riak-core-meta-data was slowing down the whole system by huge amounts.

(The change also appears to improve the visibility of updates by tens of seconds, but I don't have a precise test for this written)

riakmd

summary_overall

tcrain:
PS I would have put the meta-data changes in a different pull request, but I haven't quite finished all the pieces for it and it was delaying the pub-sub so they are pulling together.

tcrain:
Here are the latencies for the graphs above (the fall on the right side of the first graph is an artefact due to not being enough operations per time period).  The big latency seems to come from the call to get() in riak-core-meta-data, which is supposed to be a local operation so maybe we were calling it incorrectly?  The reason I was thinking to move away from riak-core-meta-data was because it uses version vectors, disk storage, and has support for more complex conflict resolution built in (all things we don't need) but it would be nice to use its tree dissemination, so maybe we can extract some parts?

mdlat

read_latencies

src/inter_dc_log_sender.erl
```diff
@@ -0,0 +1,138 @@
...
+  inter_dc_pub:broadcast(Txn),
+  Id = inter_dc_txn:last_log_opid(Txn),
+  set_timer(State1#state{last_log_id = Id}).
+
+%% @doc Return smallest snapshot time of active transactions.
+%%      No new updates with smaller timestamp will occur in future.
+get_stable_time(_Partition) ->
+    Now = inter_dc_utils:now_millisec(),
+    Now.
+%% case clocksi_vnode:get_active_txns_call(Partition) of
```
tcrain: This was commented out because I saw it was a bottleneck, so it is incorrect, I have a fix that shouldn't be a bottleneck that I will make tomorrow.

bieniusa:
@Eipifi @tcrain What is the status of this branch?

Eipifi:
New InterDC is on a different branch, the updated version will be submitted via a different pull request. This one can be closed.

Issue #149, opened 2015-09-17
multiple_dcs_test fails


Maryam81609:
In pubsub_cert branch, I cannot understand why "multiple_dcs_test" fails when it only has parallel_writes_test ("simple_replication_test", and "failure_test" are removed)?

Eipifi:
The test indeed fails after removing the simple_replication_test. **It also fails the exact same way on master branch**. It turns out that the issue is caused by wrong vectorclock comparison inside materializer_vnode.

parallel_writes_test sends 5 counter increments to 3 DCs, and expects all DCs to eventually see the value counter of 15. Instead, the DC returned the value of 5, even though remote updates already arrived and were stored in the log.

As the updates wre sent by the client, each DC handled its own updates and reached the counter value of 5. The last snapshot time for the counter was the commit time of the last increment operation.

When the client requested the value using a snapshot time (expecting to observe 15), the txn cordinator correctly waited until the stable snapshot reaches the specified snapshot time. Then, the object was materialized. The problem started here: 

https://github.com/SyncFree/antidote/blob/pubsub_cert/src/clocksi_materializer.erl#L121

The update operation from the remote DC was checked if it should be included in the new snapshot. The condition is that the incoming operation must not be included in the last snapshot, and that the operation commit time must be smaller than the snapshot time requested by the client.

The check is being done by the belongs_to_the_snapshot_op function here:

https://github.com/SyncFree/antidote/blob/pubsub_cert/src/materializer_vnode.erl#L295

The vectorclock comparison incorrectly fails because the last snapshot value at the receiving DC only holds the entry for the local DC (as only the local updates were processed so far). The vectorclock:le()/2 expects all the entries for all DCs to be in the vectorclocks given in the arguments. If there is an entry missing, the function returns false. This is why the updates from the remote DCs were not accepted and the read returned 5 instead of 15 - the belongs_to_the_snapshot_op function simply rejected those.

So why did the simple_replication_test "solve" the issue? The theory is that the simple_replication_test sent the updates from all DCs to all DCs with no concurrency involved (as it happens in multiple_dcs_test, where the updates are issued all at once). As a result, the list of the counter snapshots in the materializer might have been reordered (because these are weakly ordered to prevent the materializer from going through all of them on each read) in a way that prevented the materializer from getting the correct value. The theory is supported by the fact that putting 0 values for all DCs in counter snapshots by hand solved the issue. It's still a theory - if you have other ideas, please share them.

[...]

PR #150, opened 2015-09-18
Vectorclock operations fixed (issue #149)

Eipifi:
Solution to the problem described in https://github.com/SyncFree/antidote/issues/149.

All tests pass.

### DIFF
--- a/src/vectorclock.erl
+++ b/src/vectorclock.erl
@@ -124,91 +124,33 @@ set_clock_of_dc(DcId, Time, VectorClock) ->
 from_list(List) ->
     dict:from_list(List).
 
-eq(V1, V2) ->
-    dict:fold( fun(Dcid, Time2, Result) ->
-                       case dict:find(Dcid, V1) of
-                           {ok, Time1} ->
-                               case Time1 =:= Time2 of
-                                   true ->
-                                       Result;
-                                   false ->
-                                       false
-                               end;
-                           error ->
-                               false
-                       end
-               end,
-               true, V2).
+-spec for_all_keys(fun((non_neg_integer(), non_neg_integer()) -> boolean()), vectorclock(), vectorclock()) -> boolean().
+for_all_keys(F, V1, V2) ->
+  %% We could but do not care about duplicate DC keys - finding duplicates is not worth the effort
+  AllDCs = dict:fetch_keys(V1) ++ dict:fetch_keys(V2),
+  lists:all(fun(DC) -> F(get_clock_of_dc(DC, V1), get_clock_of_dc(DC, V2)) end, AllDCs).
 
-le(V1, V2) ->
-    dict:fold( fun(Dcid, Time2, Result) ->
-                       case dict:find(Dcid, V1) of
-                           {ok, Time1} ->
-                               case Time1 =< Time2 of
-                                   true ->
-                                       Result;
-                                   false ->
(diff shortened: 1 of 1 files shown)


src/vectorclock.erl
```diff
@@ -124,91 +124,33 @@ set_clock_of_dc(DcId, Time, VectorClock) ->
...
-                                       false
-                               end;
-                           error ->
-                               false
-                       end
-               end,
-               true, V2).
+-spec for_all_keys(fun((non_neg_integer(), non_neg_integer()) -> boolean()), vectorclock(), vectorclock()) -> boolean().
+for_all_keys(F, V1, V2) ->
+  %% We could but do not care about duplicate DC keys - finding duplicates is not worth the effort
```
bieniusa: Can you explain what you mean with "is not worth the effort"?
Shouldn't it hold that each DC only has one entry?

src/vectorclock.erl
```diff
@@ -124,91 +124,33 @@ set_clock_of_dc(DcId, Time, VectorClock) ->
...
-                                   false ->
-                                       false
-                               end;
-                           error ->
-                               Result
-                       end
-               end,
-               true, V2).
+-spec le(vectorclock(), vectorclock()) -> boolean().
+le(V1, V2) -> for_all_keys(fun(A, B) -> A =< B end, V1, V2).
```
bieniusa: You are renaming a function here!

Eipifi:
Vectorclock is a dict mapping from a DCID to a timestamp. If a given DCID key is absent, the default value is zero. When comparing two vectorclocks, you need to run the comparison predicate (<, >, == etc) for each key in both vectorclocks. We could compute a union of two sets of keys (complexity n log n), but the faster way is to just run the same predicate for all keys in the first dict, and then run the same predicate for the keys in second dict (complexity n), assuming that the predicate itself is of complexity 1.

We will compare the same entries twice, but in the end it is still faster.

src/vectorclock.erl
Eipifi: Vectorclock is a dict mapping from a DCID to a timestamp. If a given DCID key is absent, the default value is zero. When comparing two vectorclocks, you need to run the comparison predicate (<, >, == etc) for each key in both vectorclocks. We could compute a union of two sets of keys (complexity n log n), but the faster way is to just run the same predicate for all keys in the first dict, and then run the same predicate for the keys in second dict (complexity n), assuming that the predicate itself is of complexity 1.

We will compare the same entries twice, but in the end it is still faster.

tcrain: I might suggest using a sorted list (by DCID).  These would have a smaller memory footprint compared to the dict (less network traffic) and comparisons should then be 2N.  Insertions and single lookups/updates would be slower, but it seems that these are the rare cases.

I think the most common case will be in the materialiser during a read where we have to traverse a  list of operations and snapshots to find the right version to create.  The number of vector clock comparisons you have to make here is only bounded by concurrency and since this is the read path it is probably important to optimise.  The current suggestion looks to have cost 8N for a comparison plus making a copy of the keys of both lists.  While traversing two sorted lists should be 2N.  We would probably have to rework some parts of the materialiser to account for this also though.

I guess it doesn't really matter unless we start getting really big vector clocks. So maybe it is only worth looking at this if we ever find it is an issue.

PR #152, opened 2015-09-23
CRDT RGA

santialvarezcolombo:
This PR contains the implementation of the CRDT RGA (Replicated Growable Array), with its corresponding eunit and riak tests.

The implementation is based on this [paper](http://hal.upmc.fr/inria-00555588/), without any of the  optimisations discussed in this other [paper](http://dl.acm.org/citation.cfm?id=1931272), which should be future work.

The module crdt_rga includes the method *is_operation* which is intended to work once #142 is merged.


src/crdt_rga.erl
```diff
@@ -0,0 +1,288 @@
+%% @doc
```
bieniusa: Can you add the copyright notice?

src/crdt_rga.erl
```diff
@@ -0,0 +1,288 @@
...
+-spec recursive_remove(vertex(), rga(), list()) -> rga_result().
+recursive_remove(_, [], L) ->
+    {ok, L};
+recursive_remove({_, Value, UID}, [{_, Value, UID} | T], L) ->
+    {ok, lists:reverse(L) ++ [{deleted, Value, UID}] ++ T};
+recursive_remove(Vertex, [H | T], L) ->
+    recursive_remove(Vertex, T, [H | L]).
+
+%% @doc given an rga, this mehtod looks for all tombstones and removes them, returning the tombstone free rga.
+-spec purge_tombstones(rga()) -> rga_result().
```
bieniusa: Please rewrite using lists:filter !!

src/crdt_rga.erl
```diff
@@ -0,0 +1,288 @@
...
+%% It returns either the added Vertex and the new rga in case of an addRight, and the new rga in the case of a remove.
+-spec update(rga_downstream_op(), rga()) -> rga_result().
+update({addRight, RightVertex, NewVertex}, Rga) ->
+    recursive_insert(RightVertex, NewVertex, Rga, []);
+update({remove, Vertex}, Rga) ->
+    recursive_remove(Vertex, Rga, []).
+
+%% Private
+%% @doc recursively looks for the Vertex where the new element should be put to the right of.
+-spec recursive_insert(vertex(), vertex(), rga(), list()) -> rga_result().
```
bieniusa: Please rewrite in terms of take_while !!
This re-building of lists and reversing is sub-optimal.

src/crdt_rga.erl
```diff
@@ -0,0 +1,288 @@
...
+        _ ->
+            add_element({Status, Value, UID}, T, [{Status1, Value1, UID1} | L])
+    end;
+add_element(Insert, [], L) ->
+    {ok, lists:reverse([Insert | L])}.
+
+%% Private
+%% @doc recursively looks for the Vertex to be removed. Once it's found, it's marked as "deleted".
+%% The Vertex is not removed from the list, to allow adding elements to its right.
+-spec recursive_remove(vertex(), rga(), list()) -> rga_result().
```
bieniusa: Please rewrite in terms of lists:keyreplace !!

src/crdt_rga.erl
```diff
@@ -0,0 +1,288 @@
...
+    ?assertEqual(false, is_operation({remove, -1})),
+    ?assertEqual(true, is_operation({remove, 0})),
+    ?assertEqual(true, is_operation({remove, 1})),
+
+    %% invalid operation checks
+    ?assertEqual(false, is_operation({anything, 1})),
+    ?assertEqual(false, is_operation({add, 1, 4})).
+
+-endif.
+
```
bieniusa: Remove trailing lines.

src/crdt_rga.erl
santialvarezcolombo: Done!

src/crdt_rga.erl
santialvarezcolombo: Done!

src/crdt_rga.erl
santialvarezcolombo: Annette, according to the documentation of the method lists:take_while: "returns the longest prefix of the list for which all elements satisfy the predicate", but I also need the other elements which come after. In order to do it like this, I think I would also need to use lists:substract and lists:merge to get the result I want. I don't know if its better to do it this way regarding complexity. What do you think?

riak_test/oprga_test.erl
```diff
@@ -0,0 +1,92 @@
+-module(oprga_test).
+
+-export([confirm/0]).
+
+-include_lib("eunit/include/eunit.hrl").
+-define(HARNESS, (rt_config:get(rt_harness))).
+
+confirm() ->
+    [Nodes] = rt:build_clusters([3]),
```
tcrain: Maybe add these lines like was done in https://github.com/SyncFree/antidote/pull/143 to speed up the tests

    NumVNodes = rt_config:get(num_vnodes, 8),
    rt:update_app_config(all,[
        {riak_core, [{ring_creation_size, NumVNodes}]}
    ]),

Also you should probably add the lines to wait until the ring is converged since we don't have the handoff stuff implemented.

src/crdt_rga.erl
```diff
@@ -0,0 +1,298 @@
...
+-spec purge_tombstones(rga()) -> rga_result().
+purge_tombstones(Rga) ->
+    L = lists:filter(fun({Status, _, _}) ->
+        Status == ok
+    end, Rga),
+    {ok, L}.
+
+%% @doc generate a unique identifier (best-effort).
+unique() ->
+    crypto:strong_rand_bytes(20).
```
tcrain: Is this the right function to use?  Apparently it can throw exceptions and might be stronger than what we need.

How are the UIDs used?

The people at erlang solutions mentioned this
http://www.erlang.org/doc/man/erlang.html#make_ref-0
But this only checks if equal and might not work across clusters (I think).
What about a node id (atom) and increasing time?  I think atoms can be compared using <
http://erlang.org/doc/reference_manual/expressions.html#id80214

santialvarezcolombo: I copied the idea from the [crdt_orset](https://github.com/SyncFree/antidote/blob/master/src/crdt_orset.erl#L239) to generate unique IDs...

They are used to decide where to insert in case of a concurrent insertion. 

This was de easiest and simplest way to do it. Maybe we can think a better way to generate this UIDs...

santialvarezcolombo: I copied the idea from the [crdt_orset](https://github.com/SyncFree/antidote/blob/master/src/crdt_orset.erl#L239) to generate unique IDs...

They are used to decide where to insert in case of a concurrent insertion. 

This was de easiest and simplest way to do it. Maybe we can think a better way to generate this UIDs...

tcrain: What about a tuple {node(), make_ref()}?

They should be unique and comparable, I tried:

15> Id1 = {node(),make_ref()}.
{nonode@nohost,#Ref<0.0.0.97>}
16> Id2 = {node(),make_ref()}.
{nonode@nohost,#Ref<0.0.0.102>}
17> Id1 < Id2.
true
18> Id1 > Id2.
false

tcrain: Ah, there is this also:
http://www.erlang.org/doc/man/erlang.html#unique_integer-1
maybe
{node(),erlang:unique_integer([])}

is better than make_ref() because I'm not sure how the reference is stored exactly.

Issue #131, opened 2015-06-10
No easy way to turn off SI certification check


tcrain:
Right now most the the objects in antidote are CRDTs that don't need to abort in the presence of concurrent updates.  For these objects there should be no SI certification check as it will just cause unnecessary aborts and wasted CPU cycles.

Currently the only way to turn on or off the certification check it to comment it out in the code.  Some branches have it commented out and some do not.  Maybe there is a better way to do this?  For example only do certification check depending on the object?  We will have to precisely define what kind of consistency we want to guarantee though.

cmeiklejohn:
+1, we *need* this.

tcrain:
Following the mumble discussion today, the next step for this is to come up with a specification for how and when the certification check should be used.  Implementing the decided specification will also solve issue 130 ( https://github.com/SyncFree/antidote/issues/130 ).

Some initial thoughts for the specification:

- The certification check should be disabled by default as we are targeting CRDTs and causality.
- Choosing whether to certify or not could be done based on object type (for example the bounded counter would have it enabled, while the other CRDTs would not).
- Another possibility would be to give the programmer the ability to enable or disable it in application code.  This might be per object, per transaction, per workload.

marc-shapiro:
The begin_transaction construct contains a parameter "red/blue".  Blue has no certification check.  Furthermore, per CRDT type, operations are classified as red or blue (possibly depending on parameters).   A blue transaction may not contain any red operations.

						Marc

preguica:
I agree - ideally you would want to be able to specify that a transaction
should have SI check or not, but then you would like to be able to say that
the check is only done for some objects (those where you have executed one
operation that requires strong consistency).

Having said that, all this flexibility might not be that easy to code and
some compromise may be needed - e.g. whenever a strong object (e.g. bounded
counter) is used, SI check is done (independently of the executed
operation).

bieniusa:
In addition, if you offer so much flexibility, then the programming model becomes very complex.
The semantics of transactions with respect to their isolation would become non-composable, I guess.

Is this really what we want?

For example, you have the SI check for an object enabled in one txn but not in another. 
What does this mean?
Programmers are having lots of trouble understanding SI - they seem to prefer either strong (i.e. serializable) or kind of no guarantees.

	Annette

marc-shapiro:
> Le 30 juin 2015 à 18h33, Annette Bieniusa <notifications@github.com> a écrit :
> 
[...]

Issue #130, opened 2015-06-10
Certification check for SI is costly/incorrect


tcrain:
Current implementation requires storing a history of all updates and all commit records in a cache.  Most protocols avoid this by having a integer per key of the last time it was written, or if this is too costly, a hash table of integers (this can cause false aborts though).

Also the current implementation seems incorrect.  A transaction should abort if what it is updating has a newer version at the DC than what is given by the transaction's snapshot.  Right now it only takes into account updates that from the local DC and it uses the TxId for ordering, but the TxId doesn't give commit order.

tcrain:
Actually considering the part about aborting due to updates being received from other DCs is debatable.  On one hand you could say that the local transaction shouldn't abort because the update from the other DC could have just as easily arrived after the local transaction committed, thus having the same affect, especially if were are only considering certain mergeable objects.  But it is not quite the same because first it allows concurrent readers to observe different orders of the history that wouldn't be observable if the local transaction aborted and also it might matter for certain objects where the order of operations is important, which might be interesting for certain types of objects that could have "home DCs" where operations that originate at the home DC are different or are treated differently than operations that do not.  Possibly worth thinking about.

marsleezm:
Hi, what you are considering seems to relate to Parallel Snapshot Isolation(http://news.cs.nyu.edu/~jinyang/pub/walter-sosp11.pdf). What they do is when reading, a transaction only reads the latest result from local DC but updates to each key still have establish a total order among DCs, which can be achieved by certifying writes in all DCs or having a primary to serialize updates. In practice, they avoid serializing updates by only using CRDT(called c-set).

I agree that the current certification has weird semantics, because it only serializes operations in a single DC but not among DC. While on the other hand, since we are only using CRDTs, we only need it to tell if a transaction can commit or not (e.g. for bounded counter), rather than establishing a total order for writes.

tcrain:
I agree it has a similar air to PSI, but it is not quite the same because here any DC is accepting the updates, thus not admitting a total order of writes.

The current version (when fixed to use commit order instead of start time) is not quite snapshot isolation at a single DC either because an update can arrive from a different DC that invalidates your snapshot locally, but you will still commit.  But like I mentioned before this might not matter for many workloads because for most cases you eventually have the same result (but see different states in-between).

[...]

PR #153, opened 2015-09-24
Makefile targe to disable certification check

tcrain:
The idea here is to be able to disable the certification check at compile time by using the make target:
make relnocert

This way we don't have to make separate branches for benchmarking.  I guess eventually the cert check will be done dynamically deeding on the object and what the client wants.
The issues with the certification check are described here:
https://github.com/SyncFree/antidote/issues/131
https://github.com/SyncFree/antidote/issues/130


rebar.config.script
```diff
@@ -0,0 +1,13 @@
...
+case os:getenv("NO_CERTIFICATION") of
+    false ->
+        CONFIG;
+    _ ->
+        case lists:keysearch(erl_opts, 1, CONFIG) of
+            {value, {erl_opts, Opts}} ->
+                NewOpts = [NoCert | Opts],
+                lists:keyreplace(erl_opts, 1, CONFIG, {erl_opts, NewOpts});
+            false ->
+                CONFIG ++ [{erl_opts, NoCert}]
```
bieniusa: Is it necessary to append to the end? It is in general better to append at the front.

bieniusa:
Tyler, can you resolve the conflicts with master?

I would at least add another Makefile target to run the test suite under different configurations (i.e. with and without the check). And one large target that runs all the configurations.

tcrain:
If you run
make rel
It is the same as before (it includes the certification check).

If you run
make relnocert
Then when you run the tests it will only run the tests that don't require the certification check.

They both get built to the same "rel" directory, so you can only have one version at a time.
Otherwise I could make them build to different directories and configure the tests to run in both directories. Is this what you mean?

I guess the goal is to eventually be able to build antidote with different consistencies so maybe we should have a standard way of doing this.

Otherwise the rebar profiles seem like something interesting to use here, but needs rebar3
https://www.rebar3.org/docs/profiles

bieniusa:
> Am 29.09.2015 um 20:07 schrieb tcrain <notifications@github.com>:
> 
> If you run
> make rel
> It is the same as before (it includes the certification check).
> 
> If you run
> make relnocert
> Then when you run the tests it will only run the tests that don't require the certification check.
> 
> They both get built to the same "rel" directory, so you can only have one version at a time.
> Otherwise I could make them build to different directories and configure the tests to run in both directories. Is this what you mean?

Yes, having one make target that builds and tests both configuration options would be necessary at least for use in the CI server.

bieniusa:
@tcrain Please fix conflicts

tcrain:
Conflicts are resolved.

Also I created an issue https://github.com/SyncFree/antidote/issues/165 to track this until we come up with a more complete way to build and test different versions.

PR #154, opened 2015-09-28
Fix batch operations RPC 

santialvarezcolombo:
While refactoring #152 with @tcrain, we found that the batch operations in an RPC call, where executed in a different order than expected, causing a new riak_test to fail. So this PR fixes this issue and is intended to be merged before #152.


tcrain:
The problem happens when someone executes a single transaction that has multiple updates to the same key.  Subsequent reads to the key would observe the updates in the reverse order that they happened.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -226,7 +226,7 @@ perform_update(Args, Updated_partitions, Transaction, Sender) ->
                                            [{IndexNode, [{Key, Type, DownstreamRecord}]} | Updated_partitions];
                                        _ ->
                                            lists:keyreplace(IndexNode, 1, Updated_partitions,
-                                               {IndexNode, [{Key, Type, DownstreamRecord} | WriteSet]})
+                                               {IndexNode, WriteSet ++ [{Key, Type, DownstreamRecord}]})
```
bieniusa: Appending at the end is expensive. Can you also prepend?

tcrain:
The problem with the list is that we need to read/process it in the reverse order than we append to it.  This is because as updates happen during a transaction they are appended to the write set (which is stored as a list), then if a read happens these updates need to be read from the oldest first.  So either we use the expensive append, or we reverse the list before processing it for the read.  The write set also has to be reversed before storing the updates in the materialiser because the operations are stored in the materialiser using a fold left which needs the oldest operations at the head of the list (a fold right isn't used because it is more expensive).

I made a change so that we can avoid performing the reverse during the read, because the updates have to be filtered by key, and a filter is more efficient if you reverse the list during it anyway.

This bug actually extended into the materialiser because we materialise a snapshot by reading the most recent operations until we find one that was already included in the previous snapshot, and we were applying them directly as we read them, when instead they should be applied in the reverse order.  I fixed this by collecting the operations in a list, then applying them in the correct order.

src/clocksi_vnode.erl
```diff
@@ -581,15 +582,15 @@ update_materializer(DownstreamOps, Transaction, TxCommitTime) ->
...
-
+reverse_and_filter_updates_per_key(Updates, Key) ->
+    lists:foldl(fun({KeyPrime, _Type, Op}, Acc) ->
+			case KeyPrime == Key of
+			    true ->
+				[Op | Acc];
+			    false ->
+				Acc
+			end
+		end, [], Updates).
```
bieniusa: Why did you rewrite this function? The new version is equivalent, it only has a different name.

PR #159, opened 2015-10-05
Some fixes in case statements

tcrain:


src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -247,9 +249,9 @@ perform_update(Args, Updated_partitions, Transaction, Sender) ->
                         undefined ->
                             ok;
                         _ ->
-                            _Res = gen_fsm:reply(Sender, {error, Error}),
-                            {error, Error}
-                    end
+                            _Res = gen_fsm:reply(Sender, {error, Error})
+                    end,
+		    {error, Error}
```
bieniusa: Isn't the {error,Error} line now dead code??

PR #160, opened 2015-10-06
Fix certification check

marsleezm:
This branch fixes the certification check of clocksi_vnode, which was not correctly implemented and was inefficient. All riak_tests have passed.
Two tables are used to keep metadata for certification check: 
committed_tx that keeps the commit time of the last committed transaction for each key.
prepared_tx that keeps the txid of the prepared transaction and prepare_time for each key.
The certification check for a transaction, let's say T, works as follows:
1. For each key in T's write set, if T's snapshot time is smaller than the commit time of the last committed transactions for any key, T gets aborted.
2. Otherwise, if any key in T's write key has been prepared by other transaction, T is aborted. I implemented in this pessimistic way for the simplification of implementation.


src/clocksi_readitem_fsm.erl
```diff
@@ -209,20 +209,19 @@ check_clock(Key,Transaction,PreparedCache,Partition) ->
...
+    case ets:lookup(PreparedCache, Key) of
+        [] ->
+            ready;
+        [{Key, {_PreparedTxId, PrepareTime}}] ->
+            case PrepareTime =< SnapshotTime of
+                true ->
+                    {not_ready, ?SPIN_WAIT};
+                false ->
+                    ready
+            end
```
tcrain: Please keep the list here.  There can be multiple transactions being prepared for the same key at the same time, and for safety all of the times need to be checked.

src/clocksi_vnode.erl
```diff
@@ -470,7 +416,7 @@ commit(Transaction, TxCommitTime, Updates, CommittedTx, State) ->
             Transaction#transaction.vec_snapshot_time}},
     case Updates of
         [{Key, _Type, {_Op, _Param}} | _Rest] ->
-            true = ets:insert(CommittedTx, {TxId, TxCommitTime}),
+            true = ets:insert(CommittedTx, {Key, TxCommitTime}),
             LogId = log_utilities:get_logid_from_key(Key),
```
tcrain: I think you need to insert in the table for each key in the write set.

marsleezm: Yes you are right. I will fix this.

src/clocksi_vnode.erl
```diff
@@ -91,30 +90,6 @@ read_data_item(Node, TxId, Key, Type, Updates) ->
             {error, Reason}
     end.
 
-
-%% @doc Return active transactions in prepare state with their preparetime for a given key
-%% should be run from same physical node
-get_active_txns_key(Key, Partition, TableName) ->
-    case ets:info(TableName) of
-        undefined ->
```
tcrain: I would recommend keeping keeping the check for ets:info, the tests pass because at startup there is a wait loop that checks the vnodes have started and loaded the ets tables.  If the wait loop isn't run then there will be crashes in other places, for example in the benchmarks.  You can directly move this code into the readitem fsm if you want.

src/clocksi_readitem_fsm.erl
tcrain: To explain, having multiple transactions being prepared is only needed for CRDTs and not for SI because in SI with concurrent transactions being prepared on the same key, one will abort [like you have done in the certification check], but in the case of CRDTs they will both be committed.  This is something we need to take into account in the future when we have a system that decides at runtime when to perform the certification check and when not to.

tcrain:
Thanks, looks good.

Just to note, this branch solves the main issue in https://github.com/SyncFree/antidote/issues/130

bieniusa:
In this branch, the clocksi_test fails:

================ clocksi_test failure stack trace =====================
{{assertMatch_failed,
     [{module,clocksi_test},
      {line,326},
      {expression,"CommitTime"},
      {pattern,"{ ok , _ }"},
      {value,
          {badrpc,
              {'EXIT',
                  {timeout,
                      {gen_fsm,sync_send_event,
                          [<10988.1431.0>,{prepare,two_phase}]}}}}}]},
 [{clocksi_test,'-clocksi_tx_noclock_test/1-fun-2-',1,
      [{file,"riak_test/clocksi_test.erl"},{line,326}]},
  {clocksi_test,clocksi_tx_noclock_test,1,
      [{file,"riak_test/clocksi_test.erl"},{line,326}]},
  {clocksi_test,confirm,0,[{file,"riak_test/clocksi_test.erl"},{line,57}]}]}
=======================================================================

bieniusa:
Hmm, I can no longer reproduce this failed test. Probably some issue with merging from master?

marsleezm:
I just ran the tests and all passed. So yes, that’s possible..

PR #163, opened 2015-10-12
A generic transactional interface for antidote and its protocol buffer interface

deepthidevaki:
Public interface for antidote is modified to use a generic transactional interface instead of clocksi specific functions. The protocol buffer implementations is modified accordingly in SyncFree/antidote_pb and SyncFree/riak_pb repositories. 
Old rpc interface is kept as it is so that tests and benchmarks doesn't break. It will be eventually removed after all tests and benchmarks are fixed.


src/antidote.erl
```diff
@@ -36,22 +55,134 @@
...
+%% Object creation and types
+-spec create_bucket(bucket(), type()) -> {ok, bucket()}.
+create_bucket(Bucket, _Type) ->
+    %% TODO: Bucket is not currently supported
+    {ok, Bucket}.
+
+-spec create_object(key(), type(), bucket()) -> {ok, {key(), type(), bucket()}}.
+create_object(Key, Type, Bucket) ->
+    %% TODO: Object creation is not currently supported
+    {ok, {Key, Type, Bucket}}.
```
bieniusa: Shouldn't this yield a currently_not_supported error as with the other not-yet implemented features?

src/antidote.erl
```diff
@@ -36,22 +55,134 @@
...
+        {ok, {_TxId, Result, CommitTime}} ->
+            {ok, Result, CommitTime};
+        {error, Reason} -> {error, Reason}
+    end.
+
+
+%% Object creation and types
+-spec create_bucket(bucket(), type()) -> {ok, bucket()}.
+create_bucket(Bucket, _Type) ->
+    %% TODO: Bucket is not currently supported
```
bieniusa: Shouldn't this yield a currently_not_supported error as with the other not-yet implemented features?

deepthidevaki:
not_supported error is added to those methods now.

PR #166, opened 2015-10-14
Check prepared time without blocking reads

tcrain:
In antidote when there have been no updates for a certain period of time, a heartbeat is sent.  The time for the heartbeat has to have a time smaller than any transactions that might commit in the future on this partition.  Because of this, the node that sends the heartbeats has to check with the clocksi_vnode incase a transaction is being prepared that might get a smaller commit time.

Previously this was done by taking the ets table that stored the prepared transactions and changing it to a list then taking the minimum.  The problem with this is that doing this was that the ets:tab_to_list operation was blocking concurrent readers and becoming a bottleneck when there were multiple DCs heartbeats had to be sent often.  Now this information is just stored in the clocksi_vnode directly, so concurrent reads don't have to be blocked, instead writes will be blocked but only for the time it takes to read the head of the list.  Also this only happens when there have been no writes recently so this shouldn't be a problem.

### DIFF
--- a/src/clocksi_transaction_reader.erl
+++ b/src/clocksi_transaction_reader.erl
@@ -195,22 +195,11 @@ get_sorted_commit_records(Commitrecords) ->
 %% @doc Return smallest snapshot time of active transactions.
 %%      No new updates with smaller timestamp will occur in future.
 get_stable_time(Partition, Prev_stable_time) ->
-    case clocksi_vnode:get_active_txns(
-	   Partition, clocksi_vnode:get_cache_name(Partition,prepared)) of
-        {ok, Active_txns} ->
-            lists:foldl(fun({_TxId, Snapshot_time}, Min_time) ->
-                                case Min_time > Snapshot_time of
-                                    true ->
-                                        Snapshot_time;
-                                    false ->
-                                        Min_time
-                                end
-                        end,
-                        clocksi_vnode:now_microsec(erlang:now()),
-                        Active_txns);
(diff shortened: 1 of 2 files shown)


bieniusa:
When I run the riak-tests, they keep failing in a non-deterministic way.
The first time, the opbcounter-test was failing, the next time the mvreg-test.
Any idea?
```
================ mvreg_test failure stack trace =====================
{{assertEqual_failed,[{module,rt},
                      {line,618},
                      {expression,"wait_until ( F )"},
                      {expected,ok},
                      {value,{fail,false}}]},
 [{rt,'-wait_until_no_pending_changes/1-fun-3-',2,
      [{file,"src/rt.erl"},{line,618}]},
  {rt,wait_until_no_pending_changes,1,[{file,"src/rt.erl"},{line,618}]},
  {rt,maybe_wait_for_changes,1,[{file,"src/rt.erl"},{line,437}]},
  {rt,plan_and_commit,1,[{file,"src/rt.erl"},{line,403}]},
  {rt,join_cluster,1,[{file,"src/rt.erl"},{line,1015}]},
  {rt,'-build_clusters/1-lc$^0/1-0-',1,[{file,"src/rt.erl"},{line,329}]},
  {rt,build_clusters,1,[{file,"src/rt.erl"},{line,328}]},
  {mvreg_test,confirm,0,[{file,"riak_test/mvreg_test.erl"},{line,35}]}]}
```

PR #172, opened 2015-11-10
Materialiser fixes

tcrain:
This pull request aims to fix issues that could happen in the materialiser due to concurrency and cases where operations are executed before the DCs have connected fully connected and stabilised.

This pull request continues on the issues and fixes mentioned here:
https://github.com/SyncFree/antidote/issues/149
https://github.com/SyncFree/antidote/pull/150

The fix to the vector clock from https://github.com/SyncFree/antidote/pull/150 fixed certain cases, but there were still issues where operations could arrive to the materialiser from another DC with certain concurrency so that the snapshot would skip operations that should have been included.  The same could happen when garbage collecting the snapshots.  What I did to fix this was to simply assign an id number per key to each operation added to the materialiser, then when a new snapshot is generated from an old one, it will search all operations until the last one that was included by the previous snapshot.  Some unit tests have been added to check theses cases.  Note that this bug would also show up when running the inter_dc_rep_test, but only rarely, due to concurrency.  This pull request is a precursor to Michals pubsub changes as the bug is more apparent there.

### DIFF
--- a/riak_test/common.erl
+++ b/riak_test/common.erl
@@ -20,8 +20,7 @@
 -module(common).
 
 -export([clean_clusters/1,
-         setup_dc_manager/3,
-	 setup_dc_manager_no_rt/3]).
+         setup_dc_manager/3]).
 
 %% It cleans the cluster and formes a new one with the same characteristic of the input one.
 %% In case the clean_cluster parameter is set to false in the riak_test configuration file
@@ -60,29 +59,6 @@ setup_dc_manager(Clusters, Ports, Clean) ->
                                     rt:wait_until_registered(Node, inter_dc_manager),
                                     {ok, DC} = rpc:call(Node, inter_dc_manager, start_receiver,[Port]),
                                     rt:wait_until(Node,fun wait_init:check_ready/1),
-                                    Acc ++ [{Node, DC}]
-                                end, [], CombinedList),
-            lists:foreach(fun({Node, DC}) ->
-                            Sublist = lists:subtract(Heads, [{Node, DC}]),
-                            lists:foreach(fun({_, RemoteDC}) ->
-                                            ok = rpc:call(Node, inter_dc_manager, add_dc,[RemoteDC])
-                                          end, Sublist)
-                          end, Heads),
-            ok;
-        false ->
-            ok
-    end.
-
-
-setup_dc_manager_no_rt(Clusters, Ports, Clean) ->
-    case Clean of
-        true ->
-            {_, CombinedList} = lists:foldl(fun(Cluster, {Index, Result}) ->
-                                                {Index+1, Result ++ [{Cluster, lists:nth(Index, Ports)}]}
-                                            end, {1, []}, Clusters),
-            Heads = lists:foldl(fun({Cluster, Port}, Acc) ->
-                                    Node = hd(Cluster),
-                                    {ok, DC} = rpc:call(Node, inter_dc_manager, start_receiver,[Port]),
                                     Acc ++ [{Node, DC}]
                                 end, [], CombinedList),
             lists:foreach(fun({Node, DC}) ->
--- a/riak_test/common.erl
+++ b/riak_test/common.erl
@@ -20,7 +20,8 @@
 -module(common).
 
 -export([clean_clusters/1,
-         setup_dc_manager/3]).
+         setup_dc_manager/3,
+	 setup_dc_manager_no_rt/3]).
 
 %% It cleans the cluster and formes a new one with the same characteristic of the input one.
 %% In case the clean_cluster parameter is set to false in the riak_test configuration file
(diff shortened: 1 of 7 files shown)


bieniusa:
For me the clock_si_test now does not pass:

```
================ clocksi_test failure stack trace =====================
{badarg,[{erlang,hd,[{badrpc,timeout}],[]},
         {clocksi_test,clocksi_test_prepare,1,
                       [{file,"riak_test/clocksi_test.erl"},{line,262}]},
         {clocksi_test,confirm,0,
                       [{file,"riak_test/clocksi_test.erl"},{line,69}]}]}
=======================================================================
```

bieniusa:
Or (non-deterministically) some other tests:

```
19:17:26.180 [error] 
================ opcounter_test failure stack trace =====================
{{assertMatch_failed,
     [{module,opcounter_test},
      {line,226},
      {expression,"Result"},
      {pattern,"{ ok , _ }"},
      {value,
          {badrpc,
              {'EXIT',
                  {timeout,
                      {gen_fsm,sync_send_event,[<11031.1186.0>,execute]}}}}}]},
 [{opcounter_test,'-clocksi_single_key_update_read_test/1-fun-1-',1,
      [{file,"riak_test/opcounter_test.erl"},{line,226}]},
  {opcounter_test,clocksi_single_key_update_read_test,1,
      [{file,"riak_test/opcounter_test.erl"},{line,226}]},
  {opcounter_test,confirm,0,
      [{file,"riak_test/opcounter_test.erl"},{line,60}]}]}
=========================================================================
```

tcrain:
In the clock si test the timeout you are seeing is from the call on line 261 to the log utilities module
https://github.com/SyncFree/antidote/blob/fix-mat-ss-gc/riak_test/clocksi_test.erl#L261

I haven't touched this code and the only real work it does is call riak core and I don't know what would be happening in riak core to cause the timeout.  One possibility could be if the changes I added would slow down the rest of the system, but it seems unlikely since the code I added wouldn't be running during this call.

For the error in opcounter_test, I just made a new commit for testing where I increased the timeout in the call to the transaction coordinator, so maybe now we can see if something is blocking somewhere or at least get a more precise error.  Unfortunately I am unable to reproduce the timeouts so if you could please try running the tests again on the updated commit.

It seems that part of the problem might be due to the general slowness of the testing suite, but it is hard to know for sure.  Especially since sometimes the tests run really fast, and other times they won't even start up and I get errors like this during initialisation and other times the tests just stall during ring creation:
11:53:48.099 [error] 
================ opcounter_test failure stack trace =====================
{{case_clause,
     {badrpc,
         {'EXIT',
             {noproc,
                 {gen_server,call,
                     [{riak_core_claimant,'dev1@127.0.0.1'},
                      plan,infinity]}}}}},
 [{rt,plan_and_commit,1,[{file,"src/rt.erl"},{line,399}]},
  {rt,join_cluster,1,[{file,"src/rt.erl"},{line,1015}]},
  {rt,'-build_clusters/1-lc$^0/1-0-',1,[{file,"src/rt.erl"},{line,329}]},
  {rt,build_clusters,1,[{file,"src/rt.erl"},{line,328}]},
  {opcounter_test,confirm,0,
      [{file,"riak_test/opcounter_test.erl"},{line,43}]}]}
=========================================================================

Or I get this repeating "forever" (until I give up):

13:06:33.890 [info] planning and commiting cluster join
13:06:33.891 [info] plan: ring not ready
13:06:34.499 [info] planning and commiting cluster join
13:06:34.500 [info] plan: ring not ready
13:06:35.107 [info] planning and commiting cluster join
13:06:35.107 [info] plan: ring not ready



One comment is that while there are likely plenty of bugs in the code I am writing, things are always much more difficult to debug then I would like given the un-documentedness and unpredictability of the riak-core base.  We see it on the test suite running on a single machine, but it gets even more difficult when I am doing lager tests with multiple machines and DCs.

The only riak-core functionality we really rely on currently is the consistent-hashing, so I might suggest we support a standalone version of antidote without riak-core (maybe through a compile time option).  This would just be for static deployments where we fix the layout of gen-fsms on the servers at start time.  This way would could more easily fix/test/optimise antidote.  Then for more realistic/dynamic deployments we could use riak core.

bieniusa:
Regarding the clocksi timeout:
It times out immedieatly it seems?
```
15:19:36.711 [info] Test prepare started
15:19:36.713 [error] Error in process <0.106.0> on node 'riak_test@127.0.0.1' with exit value: {badarg,[{erlang,hd,[{badrpc,timeout}],[]},{clocksi_test,clocksi_test_prepare,1,[{file,"riak_test/clocksi_test.erl"},{line,262}]},{clocksi_test,confirm,0,[{file,"riak_test/clocksi_test.erl"},{line,69}]}]}
```

tcrain:
Hi, I fixed an issue where during startup an ets table might be created before one with the same name had be destroyed.  It might be related to the issue you are seeing.  All tests are passing on my machine.

Maybe also try removing entirely your deps directory and your ~/rt/antidote/current/dev/dev* directories.  Sometimes I have to do this when I am frequently changing branches.

bieniusa:
It now also works for me.

Issue #151, opened 2015-09-22
Suggestion to replace erlang:now with os:timestamp


tcrain:
We use erlang:now in several places on the critical path of the client operations.  Apparently it uses a global lock to ensure that every call to this gets an increasing timestamp.  This could be a bottleneck.

Suggestion to replace this with os:timestamp.  With this different calls should be able to get the same timestamp, or concurrent processes might be able to read out of order.  For this I think we have to be careful when creating new transactions and when calculating the stable time per partition, but it should be solvable.

Eipifi:
erlang:now/0 is deprecated in erlang documentation, so we should change it anyway.
http://www.erlang.org/doc/man/erlang.html#now-0

tcrain:
Ah nice.  Maybe we can use this then?
http://www.erlang.org/doc/man/erlang.html#monotonic_time-0
Since its not strictly increasing, we might calculate the stable snapshot using < instead of =< for example.  I think we already do this?

Also maybe we should ensure transaction ids are unique.  Right now they are {time, pid}, but the same process could make two transactions with the same time if it is faster than time increases!

tcrain:
Oh no, maybe this instead:
http://www.erlang.org/doc/man/erlang.html#system_time-0

The previous one only gives an offset from some random point in time.
And we have to choose the correct time warp option I guess:
http://www.erlang.org/doc/apps/erts/time_correction.html#Time_Warp_Modes

tcrain:
I ran a benchmark just to check with 3 nodes, and it showed about a 6-7% performance increase by using os:timestamp over erlang:now.  I guess this would be more significant with more machines.

The problem is that os:timestamp isn't safe for the clocksi protocol.  Unfortunately the functions that would be safe as described my previous post are not available in the version of erlang we are currently using.  They are available in OTP 18 as described here:
http://www.erlang.org/doc/apps/erts/time_correction.html

Actually here it also says 
"However, you are strongly encouraged to use the new API instead of the old API based on erlang:now/0. erlang:now/0 has been deprecated since it is and forever will be a scalability bottleneck."

So maybe we should upgrade.

bieniusa:
The issue with upgrading is in the dependency with riak-core.

So, upgrading would be non-trivial (unless we get away from riak core).

aletomsic:
Maybe upgrading riak core to comply with erlang v.18 would not be that hard.
[...]

PR #173, opened 2015-12-01
config for erlang:now() vs os:timestamp()

tcrain:
This relates to the issue mentioned here
https://github.com/SyncFree/antidote/issues/151

erlang:now() is a well known scalability bottleneck, which is what we are currently using for clock si
os:timestamp() should scale, but can go backwards which makes clock si unsafe (rare but possible)

This branch is a temporary fix allowing you to use one or the other through a define in antidote.hrl

Eventually this will be fixed by upgrading to a new version of erlang that has the new timekeeping.

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -39,6 +39,10 @@
 %% At commit, if this is set to true, the logging vnode
 %% will ensure that the transaction record is written to disk
 -define(SYNC_LOG, false).
+%% Uncomment the following line to use erlang:now()
+%% Otherwise os:timestamp() is used which can go backwards
+%% which is unsafe for clock-si
+%% -define(SAFE_TIME, true).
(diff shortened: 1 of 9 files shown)


marc-shapiro:
Why not just always use os timestamp and add a check for clock going backwards?

						Marc

tcrain:
Since the readers and writers are multi-threaded, then we would need an extra amount of state to share between them about what clock they have seen, putting an extra serialisation point either through messages or an ets table.  This is probably a perfectly acceptable solution and better than using erlang:now(), but I would guess that the new implementation of erlang's time module handles this much better at a lower level and it keeps the code cleaner, but this is all just speculation of course.

bieniusa:
I would for the moment merge this patch, but suggest using the "safe" and correct, alas slow version as default.

PR #174, opened 2015-12-04
New interDC

Eipifi:
This pull request introduces ZeroMQ-based interDC layer.

The change removes the single receiver per DC (bottleneck, possible single point of failure). It also provides a way to retrieve lost updates as needed, and makes it possible for DCs to join and leave the network. The new interDC layer uses ZeroMQ as the communication library. It can be thought of as an abstraction over TCP and does not involve any setup, configuration or central brokers. The deployment process is effectively the same as in the previous interDC layer. The only notable difference is data flow direction — previously nodes had to open a receiver socket (input), now nodes open a publisher socket (output).

**Definitions**

*Publisher* - a gen_server existing in each node inside a DC. Broadcasts transactions committed on vnodes using a PUB zeroMQ socket.
*Subscriber* - a gen_server that listens to transactions sent by publishers through SUB zeroMQ sockets. Once a transaction is received, it is passed on to the apropriate buffer and eventually stored in the log.
*Filtering* - a feature of ZeroMQ that limits pub/sub traffic by enabling subscriber sockets to listen to messages with a defined prefix. Filtering happens on the sender (publisher) side. 
*Local publisher* - from the prerspecive of a vnode, a publisher residing on the same physical machine.

**Connecting DCs**

Each DC provides its descriptor — a structure that lists all the information necessary to connect to the DC. A descriptor can be obtained from a DC by calling inter_dc_manager:get_descriptor() on any of the nodes in a DC. In order to subscribe to another DC, a function inter_dc_manager:observe_dc(Descriptor) can be used. The connection will be reestablished if the observed DC crashes and comes back.

**Transaction structure**

    -record(interdc_txn, {
     dcid :: dcid(),
     partition :: partition_id(),
     prev_log_opid :: log_opid()
     snapshot :: snapshot_time(),
     timestamp :: non_neg_integer(),
     operations :: [operation()]
    }).

 Each operation contains the payload (operation type, arguments, etc.) and the operation id (a sequence number in the partition log). Each partition maintains its own log and assigns subsequent opids to new operations. Clients subscribing to transactions from a particular partition can track these identifiers and detect gaps.   

**Transporting transactions**

The contents of the logs are tracked by inter_dc_log_sender vnode. Once a complete transaction is stored in the log, it is sent to the local publisher. Publishers send only the transactions committed locally, not the transactions received from other DCs.

Each node inside the DC hosts its own publisher, and makes it available for the local vnodes. Each vnode uses a publisher socket only on the node (machine) that hosts it. If the location of the vnode changes (due to a failure of the hosting node), vnodes switch to other publisher. Therefore the publisher socket used by the partition may change over time — in order for the client to ensure that all updates are received, the client must subscribe to all publishers in a DC.

ZeroMQ allows subscribers to filter the messages by prefix. We use that feature by appending the partition identifier (160 bits) to every transaction. This way, nodes can filter incoming traffic to only receive a subset of needed transactions. This way we eliminate the bottleneck of a single receiver per DC, and also remove the possible single point of failure.

Each transaction contains one or more update operations, one prepare and one commit operation. Each operation is sent along with its opid (operation identifier). As these numbers increase monotonically by one, clients can verify if no messages were lost. If a gap in incoming operation ids is detected, clients can use a separate connection to fetch the missing transactions from the log using a separate connection. This is a rare behaviour and it is expected to be used only during connection loss or DC failure. This approach enables DCs to redownload missed messages in case of a DC crash, and eliminates the need for buffering on the sender side.

**Receiving transactions**

Each DC subscribes to all publishers in other DCs. As the transactions are received, the operation ids are checked for missed messages. Once the reliable delivery is ensured, the transactions are checked for their causal dependencies — each transaction is applied once its vector clock is less than or equal to the stable DC snapshot. The transaction is then stored in the log. If the dependencies are not fulfilled, the transactions are buffered in order. Once the clock increases, the buffer contents are rechecked.

**Heartbeat messages**

In order to ensure eventual delivery of all transactions, each partition sends a heartbeat message if no transaction is committed in a defined period of time. These heartbeat messages are structurally identical to normal transactions, but do not hold any operations. The heartbeat messages also contain the current timestamp of the partition, so remote DCs can use it to recalculate the stable snapshot.

The benchmark results for pubsub vs previous interDC are coming soon.


rebar.config
```diff
@@ -6,6 +6,7 @@
     {riak_dt, ".*", {git, "git://github.com/aletomsic/riak_dt", {tag, "type_check_antidote"}}},
     {riak_api, ".*", {git, "git://github.com/basho/riak_api", {tag, "2.0.2"}}},
     {riak_pb, ".*", {git, "git://github.com/syncfree/riak_pb", {tag, "new_interface"}}},
+    {erlzmq, ".", {git, "git://github.com/tcrain/erlzmq2", "master"}},
```
bieniusa: I would prefer some direct dependency on ZMQ. 
Does Tyler's version have many changes?

Eipifi: The only change introduced by Tyler is the following one: https://github.com/tcrain/erlzmq2/commit/95bb40064d3d7b252c57987f23fd67a870028db1

This was done because the default tarball address was unreachable from grid5k. I do not know if the issue still persists - we've had lots of problems with blocked traffic on grid5k.

bieniusa:
```
14:17:39.698 [error] 
================ log_handoff_test failure stack trace =====================
{{assertEqual_failed,[{module,rt},
                      {line,618},
                      {expression,"wait_until ( F )"},
                      {expected,ok},
                      {value,{fail,false}}]},
 [{rt,'-wait_until_no_pending_changes/1-fun-3-',2,
      [{file,"src/rt.erl"},{line,618}]},
  {rt,wait_until_no_pending_changes,1,[{file,"src/rt.erl"},{line,618}]},
  {log_handoff_test,test_handoff,3,
                    [{file,"riak_test/log_handoff_test.erl"},{line,81}]},
  {lists,foreach,2,[{file,"lists.erl"},{line,1323}]},
  {log_handoff_test,confirm,0,
                    [{file,"riak_test/log_handoff_test.erl"},{line,57}]}]}
===========================================================================
```

rebar.config
tcrain: I can make a separate config for grid5000 to keep the master branch clean.

tcrain:
If you are using the same riak test config as me, this is happening after a 600 second timeout waiting for ownership transfers in the ring to be completed (though I have yet to run into this specific timeout in the log handoff test on my machine).

If you don't see any other errors and it doesn't fail every time it might be related to the issue we see sometimes with the ring not joining in the beginning of the tests, an issue that I am planning to debug as soon as I have some time.  In that case my guess is that it isn't specific to this branch, but maybe there is some change done within this branch makes the issue more apparent.

Thats all I know for now, but fixing the ring not joining is something I want to look into as soon as I can because it is popping up a lot now.  If anyone has any other ideas or wants to look into it as well I can try to help.

src/dc_utilities.erl
```diff
@@ -18,14 +18,128 @@
...
+    ensure_all_vnodes_running/1,
+    ensure_all_vnodes_running_master/1,
+    get_partitions_num/0,
+    check_staleness/0]).
 
 get_my_dc_id() ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     riak_core_ring:cluster_name(Ring).
 
+get_my_dc_nodes() ->
```
bieniusa: Please add the Dialyzer specs for all exported functions.

src/inter_dc_dep_vnode.erl
```diff
@@ -0,0 +1,230 @@
...
+      type = Type,
+      op_param = Op,
+      snapshot_time = SnapshotTime,
+      commit_time = {DCID, CommitTime},
+      txid =  LogRecord#log_record.tx_id
+    }
+  end, inter_dc_txn:ops_by_type(Txn, update)).
+
+now_milisec() ->
+  {MegaSecs, Secs, MicroSecs} = os:timestamp(),
```
bieniusa: Should use dc_utiliities:now() for compatibility with other parts of the code (see https://github.com/SyncFree/antidote/pull/173 )

src/inter_dc_manager.erl
```diff
@@ -17,113 +17,110 @@
...
+  get_descriptor/0,
+  observe_dc/1,
+  observe_dc_sync/1,
+  observe/1,
+  observe_dcs/1,
+  observe_dcs_sync/1,
+  forget_dc/1,
+  forget_dcs/1]).
+
+-spec get_descriptor() -> {ok, #descriptor{}}.
```
bieniusa: Misses short documentation of exported functions.

src/inter_dc_utils.erl
```diff
@@ -17,25 +17,14 @@
...
-
-
-init([]) ->
-    Worker = {inter_dc_communication_fsm,
-              {inter_dc_communication_fsm, start_link, []},
-              transient, 5000, worker, [inter_dc_communication_fsm]},
-    {ok, {{simple_one_for_one, 5, 10}, [Worker]}}.
+-spec now_millisec() -> non_neg_integer().
+now_millisec() ->
+  {MegaSecs, Secs, MicroSecs} = erlang:now(),
```
bieniusa: Same as before: use dc_utilities:now()

src/vectorclock.erl
```diff
@@ -130,6 +104,15 @@ set_clock_of_dc(DcId, Time, VectorClock) ->
 from_list(List) ->
     dict:from_list(List).
 
+max([]) -> dict:new();
```
bieniusa: max() and min() need unit tests.

aletomsic: Hi, just to know. This has not been merged, right?
I want to include it in the eventually consistent branch.

Eipifi: Not yet. I guess you can just copy-paste to your branch for now.

bieniusa: You mean the test or the code? Or both?

Eipifi: I talked with Alejandro, I thought he referred to the min/max functions, but he meant the whole branch. So never mind.

aletomsic:
When merging with my eventual consistency branch, should I use this branch or pubsub? I guess this one, right?

Alejandro Tomsic

bieniusa:
Merge with master.


Annette Bieniusa
TU Kaiserslautern
FB Informatik
PO Box 3049
67653 Kaiserslautern

bieniusa@cs.uni-kl.de <mailto:bieniusa@cs.uni-kl.de>

Office:Building 34, Room 416
Phone:+49 (0) 631 205 3958
Fax:+49 (0) 631 205 3420

Issue #168, opened 2015-10-27
FSM and Supervisor per transaction and fast path for static transactions in protocol buffer


tcrain:
This issue was brought up in the code review by ESL, but so we track it here:

Currently every transaction starts an erlang fsm + sup when it starts then closes when the transaction commits, this is expensive.  One solution might be to start an fsm for each client connection through the protocol buffer.  This fsm will then be responsible for all transactions this client executes and can keep some state for the client if we want.

For a quick comparison I ran a benchmark on my local machine.  The first is using a fast path for single reads and writes that doesn't start an fsm per transaction (this is what I have been using for benchmarking up to now).  The second is also doing single item updates but is starting the fsm per transaction.  [Note the performance levels out faster when running on a dedicated server vs my local machine, but it shows the same trend]

A second thing to mention is that as far as I understand we don't have a fast path in the protocol buffer for static transactions, so currently they require 3 client/server rounds (begin tx/perform ops/commit) when they only need 1.  It might be worth fixing this at the same time as fixing the fsm issue.

![summary](https://cloud.githubusercontent.com/assets/5465025/10765704/51c0b348-7cd4-11e5-9f73-ea23f7033a5b.png)

![summary](https://cloud.githubusercontent.com/assets/5465025/10765714/60bd39e8-7cd4-11e5-8425-d8ca13b02243.png)

tcrain:
This issue was addressed in
https://github.com/SyncFree/antidote/pull/175
https://github.com/SyncFree/antidote/pull/184

PR #175, opened 2015-12-07
Interface fix singlekey

deepthidevaki:
This pull request add fast path single key read/update into the protocol buffer. This address the second part of issue #168.


bieniusa:
The specification for antidote:read/2 states that the function might also return {'error',atom()} but the inferred return is {'error',{'type_check',_}} | {'ok',_}

deepthidevaki:
Fixed it.

PR #176, opened 2015-12-10
Gentlerain config

deepthidevaki:
This pull request allows configure clocksi or gentlerain protocol during build.
- make rel - (default) uses clocksi
- make relgentlerain - uses gentlerain
The interfaces are same for both protocols. So no need to change clients. There is no support for interactive transactions and write only transactions in original gentlerain. So in our implementation, there is no clear semantics for interactive and write only transactions, but the interface supports them.


bieniusa:
@deepthidevaki Can you check how I did the configuration now for the certification check and mimic this for the protocol config?

deepthidevaki:
Transactiona protocols can now be configured using application config parameters.

aletomsic:
Hi Deepthi,

Could you please summarise how you implemented the GentleRain protocol in the codebase?
Or, if you have it explains somewhere, point me to it?
I am starting to implement a different snapshotting algorithm that should be more similar to GentleRain than antidote (at least, at the beginning), and would like to depart from there.
Thanks!

PR #181, opened 2015-12-16
Eiger interdc

marsleezm:
Minor change for adapting to pb interface.


marsleezm:
Should compare with eiger branch.

Issue #83, opened 2014-11-25
Improve Materialiser cache and its GC mechanism


aletomsic:
Current Implementation:

The are two per-key caches:
an operation cache where the most recent operations are stored, and
a snapshot cache which stores the latest materialised views (snapshots) of a  CRDT.
When the materialiser receives an operation, it stores in its operation cache. When it receives a read request, it checks whether there are suitable materialized views available in the snapshot cache and whether there are operations to be applied in the operation cache. If so, it materializes a new version of the CRDT that satisfies the request by applying the potentially missing operations to the corresponding snapshot, thus creating a new snapshot. In the event that cached snapshots or operations exist for a requested key, but none satisfying the read request parameters (i.e. snapshot time), the materialiser replies with an {error, no_snapshot} message.

GC
As there are object views (snapshots) and operations, the growing number of any of them needs to trigger the GC mechanism. 
1. Snapshot (or read-triggered) GC: 
     In case the number of snapshots stored for a key grows up to a value “snapshot_threshold”, keep just the latest “snapshot_min” snapshots. Proposed initial values (for no particular reason) are 10 and 2 respectively. Note that the snapshot generation is done on reads, that’s why this mechanism is triggered only on reads.
2. Operation (or write triggered) GC:
      In case the number of update operations stored for a key grows up to a number “ops_threshold”, keep just the latest “ops_min” operations by creating a snapshot of the object applying and then removing the oldest ops_threshold-ops_min ops. Proposed an initial values (for no particular reason are 50 and 5 respectively). Note that the snapshot generation could possibly trigger the Snapshot GC mechanism. 

Please contribute to the following with your ideas:

Future
Caching
The idea is to generate some policies to define how each key will be cached, depending on its type or maybe its size. It might be even possible to change policy dynamically, e.g., according to an element’s size. The goal is to achieve a more efficient handling of the cache, in both terms of size and speed on handling requests.
There should be an implementation of the materialiser read and update functions for each policy.  
NoOps: for an object cached respecting this policy, there will be no operations stored, only snapshots. This policy could be particularly handy for CRDTs which state is small compared to the size of an operation: e.g. counters, where the cost in terms of size of storing the value (a snapshot) is similar to storing the operation. In this case, for each update, a snapshot should be created. On read, the snapshots would be ready to serve requests. This would improve both size and read time for counter/like CRDTs.
[...]

PR #185, opened 2015-12-16
materialiser ets hacks

tcrain:
This deals with issue
https://github.com/SyncFree/antidote/issues/83

This addresses two major issues with the materialiser scalability, without changing the functionality.

The first deals with when a new op is added.  Before we had to load the list of ops from the ets table, make a new list with the op appended, then rewrite the entire new list back into ets.  Now it just destructively updates the op in the ets table.

The second deals with workloads that have DC locality, for example where one DC would read and update a key often while another DC would rarely access this key.  In this case at the DC where the object was rarely accessed the list of operations might overflow since garbage collection was only triggered by reads.  Now the GC is forced to trigger before this happens.

There are still many things to be done with the materialiser, but these were two main issues that could be addressed without major refactoring.

### DIFF
--- a/src/materializer_vnode.erl
+++ b/src/materializer_vnode.erl
@@ -26,7 +26,7 @@
 
 
 -define(SNAPSHOT_THRESHOLD, 10).
--define(SNAPSHOT_MIN, 5).
+-define(SNAPSHOT_MIN, 3).
 -define(OPS_THRESHOLD, 50).
 
 -ifdef(TEST).
@@ -159,7 +159,7 @@ handle_command({update, Key, DownstreamOp}, _Sender,
 
 handle_command({store_ss, Key, Snapshot, CommitTime}, _Sender,
                State = #state{ops_cache = OpsCache, snapshot_cache=SnapshotCache})->
-    internal_store_ss(Key,Snapshot,CommitTime,OpsCache,SnapshotCache),
+    internal_store_ss(Key,Snapshot,CommitTime,OpsCache,SnapshotCache,false),
     {noreply, State};
 
 
@@ -218,8 +218,8 @@ terminate(_Reason, _State=#state{ops_cache=OpsCache,snapshot_cache=SnapshotCache
 
 %%---------------- Internal Functions -------------------%%
 
--spec internal_store_ss(key(), snapshot(), snapshot_time(), cache_id(), cache_id()) -> true.
-internal_store_ss(Key,Snapshot,CommitTime,OpsCache,SnapshotCache) ->
+-spec internal_store_ss(key(), snapshot(), snapshot_time(), cache_id(), cache_id(),boolean()) -> true.
+internal_store_ss(Key,Snapshot,CommitTime,OpsCache,SnapshotCache,ShouldGc) ->
     SnapshotDict = case ets:lookup(SnapshotCache, Key) of
 		       [] ->
 			   vector_orddict:new();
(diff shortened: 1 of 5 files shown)


src/materializer_vnode.erl
```diff
@@ -281,13 +295,13 @@ internal_read(Key, Type, MinSnapshotTime, TxId, OpsCache, SnapshotCache) ->
 			ignore ->
 			    {ok, Snapshot};
 			_ ->
-			    case NewSS and IsFirst of
+			    case (NewSS and IsFirst) or ShouldGc of
```
bieniusa: Should this be an "orelse"? and an "andalso"?

src/materializer_vnode.erl
```diff
@@ -314,27 +328,28 @@ belongs_to_snapshot_op(SSTime, {OpDc,OpCommitTime}, OpSs) ->
 %% @doc Operation to insert a Snapshot in the cache and start
 %%      Garbage collection triggered by reads.
 -spec snapshot_insert_gc(key(), vector_orddict:vector_orddict(),
-                         cache_id(),cache_id() ) -> true.
-snapshot_insert_gc(Key, SnapshotDict, SnapshotCache, OpsCache)->
+                         cache_id(),cache_id(),boolean()) -> true.
+snapshot_insert_gc(Key, SnapshotDict, SnapshotCache, OpsCache,ShouldGc)->
     %% Should check op size here also, when run from op gc
-    case (vector_orddict:size(SnapshotDict))>=?SNAPSHOT_THRESHOLD of
+    case ((vector_orddict:size(SnapshotDict))>=?SNAPSHOT_THRESHOLD) or ShouldGc of
```
bieniusa: Also here "orelse"?

src/meta_data_manager.erl
```diff
@@ -74,15 +74,15 @@ init([Name]) ->
     {ok, #state{table=Table}}.
 
 handle_cast({update_meta_data, Name, NodeId, Dict}, State) ->
-    true = ets:insert(meta_data_sender:get_name(Name,?REMOTE_META_TABLE_NAME), {NodeId, Dict}),
```
bieniusa: I guess we should keep this pattern matching here, as some form of assertion.

src/materializer_vnode.erl
```diff
@@ -314,27 +328,28 @@ belongs_to_snapshot_op(SSTime, {OpDc,OpCommitTime}, OpSs) ->
...
+					[] ->
+					    {0, 0, []};
+					[Tuple] ->
+					    [Key,Length1,OpId1|Ops] = tuple_to_list(Tuple),
+					    {Length1, OpId1, sub_reverse(Ops,Length1,[])}
+				    end,
             {NewLength,PrunedOps}=prune_ops({Length,OpsDict}, CommitTime),
             ets:insert(SnapshotCache, {Key, PrunedSnapshots}),
-            true = ets:insert(OpsCache, {Key, {NewLength,PrunedOps}});
+            true = ets:insert(OpsCache, list_to_tuple([Key,NewLength,OpId|lists:reverse(PrunedOps) ++ lists:seq(NewLength+1,?OPS_THRESHOLD)]));
```
bieniusa: These lists are very short, i.e. less than 10 elements, right?

tcrain:
Ok, please review again.  I made changes so we avoid extra costs when constructing the lists and tuples during reads and GC.

The overall idea here is that if we store the operations in the ets table in a tuple per key instead of a list, then we can use a much more efficient way of adding new operations to the cache as they arrive.  We can directly update a single element in tuple in an ets table without modifying loading or modifying anything else.  Before we were making new copies of the entire list just to update a single op, then rewriting the entire list back in the table, which not only meant rewriting the entire list of operations for that key, but also invalidating the nearby pointers of the data structure as they would point to an entirely new object.

PR #191, opened 2016-01-13
Static tx parallel

marsleezm:
This branch makes static transaction coordinator to send read/write in parallel.


bieniusa:
@marsleezm This branch has dialyzer issues and needs to have merge conflicts resolved. Can you take care of this?

marsleezm:
I can not run dialyzer on my machine.. Will finish the fix by Monday.

marsleezm:
The dialyzer warning is fixed.

bieniusa:
Test 1 of the clocksi_tests is consistently blocking, also on Jenkins. As are gr_test ,inter_dc_repl_test ,logging_vnode_test ,multiple_dcs_test ,opcounter_test ,oprga_test ,pb_client_test.

marsleezm:
Fixed and all tests passed.

tcrain:
Looks good for read transactions.

I have some suggestions for the updates though, but can save them for a future pull request:
If I understand correctly the writes are still done one at a time, the issue here being that within the update a read is performed, so it still needs to wait for a round trip for each update (2 if we include the trip to the logging vnode).

A suggestion then is to separate the transaction into two phases, first the read phase where all reads are performed in parallel, including the reads for the keys that will be updated.  Then the update phase, in this phase the cached values from the already completed reads are used, so generating the downstream is a local operation.  We could also send the updates to the logging vnodes in parallel.  What do you think?

angbrav:
I did once something similar for interactive transactions which I do not really know whether it was finally merged.

Best,
Manuel

bieniusa:
I will keep this branch till Manuel has checked this feature.

aletomsic:
I had done the same, for writes, but none of them were, so that should be put as pending, again.

bieniusa:
Pull requests??

aletomsic:
If I remember correctly, they had been merged together to some benchmarking branch, that, for some, reason, was never merged to master.
Now those branches are old and conflictive to merge.

tcrain:
To not block the work, do you think we should try to merge your old branch, or is it better to redo them?

aletomsic:
I should take a look to evaluate, I will when I finish the NMSI implementation.

bieniusa:
So, you need to refactor them out again, I guess?

	Annette

> Am 25.02.2016 um 11:24 schrieb Alejandro Zlatko Tomsic <notifications@github.com>:
> 
> If I remember correctly, they had been merged together to some benchmarking branch, that, for some, reason, was never merged to master.
> Now those branches are old and conflictive to merge.
> 
>

aletomsic:
Yes.

Alejandro Tomsic

PR #192, opened 2016-01-18
Fix abort

tcrain:
This fixes two problems that were making the certification check benchmark crash.

The first is when the tx coordinator sends the abort command to the clocksi vnode, the vnode sends back an ack ( https://github.com/SyncFree/antidote/blob/master/src/clocksi_vnode.erl#L366 ), but the coordinator wasn't expecting it, causing it to crash.  Now it waits for the ack to know that the abort was successful.

The second was the execution of a static tx made up of updates was expecting a reply of either ok or error ( https://github.com/SyncFree/antidote/blob/master/src/antidote.erl#L169 ), but the tx coordinator could return ok, error, or abort, now it returns either ok or error, but in the information about the error it says it was an abort.

Actually this brings up another issue, some of the functions in the antidote.erl transaction interface return ok, error, or abort, while others just return ok or error (with abort being a possible reason for the error included with the error).  Shouldn't we choose one of these and stick with it, I like the second option personally?  But maybe this would break the protocol buffer as well?

### DIFF
--- a/src/antidote.erl
+++ b/src/antidote.erl
@@ -307,7 +307,12 @@ clocksi_execute_tx(Clock, Operations, UpdateClock, KeepAlive) ->
 			       ok = gen_fsm:send_event(TxPid, {start_tx, self(), Clock, Operations, UpdateClock}),
 			       TxPid
 		       end,
-	    gen_fsm:sync_send_event(CoordPid, execute, ?OP_TIMEOUT)
+	    case gen_fsm:sync_send_event(CoordPid, execute, ?OP_TIMEOUT) of
+		{aborted, Info} ->
+		    {error, {aborted, Info}};
+		Other ->
+		    Other
+	    end
     end.
 
 clocksi_execute_tx(Clock, Operations, UpdateClock) ->
--- a/src/clocksi_interactive_tx_coord_fsm.erl
+++ b/src/clocksi_interactive_tx_coord_fsm.erl
@@ -79,6 +79,7 @@
     committing_single/3,
     committing/3,
     receive_committed/2,
+    receive_aborted/2,
     abort/1,
     abort/2,
     perform_singleitem_read/2,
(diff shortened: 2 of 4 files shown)


marc-shapiro:
> Le 18 janv. 2016 à 13:14, tcrain <notifications@github.com> a écrit :
> 
> This fixes two problems that were making the certification check benchmark crash.
> 
> The first is when the tx coordinator sends the abort command to the clocksi vnode, the vnode sends back an ack ( https://github.com/SyncFree/antidote/blob/master/src/clocksi_vnode.erl#L366 <https://github.com/SyncFree/antidote/blob/master/src/clocksi_vnode.erl#L366> ), but the coordinator wasn't expecting it, causing it to crash. Now it waits for the ack to know that the abort was successful
> 
I wonder why: is this logically necessary?  Suppose something goes wrong and the ack is lost, is the abort aborted???

Let me reiterate yet one more time: the fate of a transaction logically depends ONLY on unanimous "prepare".  Once all transaction participants have recorded the prepare message in their log, the transaction IS committed, and all entities in the system must reach this same conclusion no matter what happens.  If there is a state where not all participants have recorded prepare, and one knows for sure that this will not happen in the future (e.g. a participant is known to have crashed) then everyone must conclude "abort".  The abort message is an optimisation, and the termination should take place even if participants crash or messages get lost.

					Marc

tcrain:
I agree with this, but this pull request is only to fix a crash where something was not expecting a message, but a message was being sent. So now it doesn't crash during normal execution.

Again, there is no recovery or fixing lost messages anywhere and I didn't implement it here, this has been discussed over and over and is something that is supposed to be considered when we start restructuring the supervisors and start fixing the fault tolerance.

Instead of changing the logic here I just fixed the crash.  I didn't want to make little hacks to ensure "sort of fault tolerance" that would be replaced later anyway. Instead, when someone does do the work to make Antidote fault tolerant they should examine the whole structure and make a valid system.

It would be great if someone took the time to look into Antidote and design a proper fault tolerance system (a unanimous prepare is just a tiny part of this), ensuring both correctness and progress on failure, crash and recovery and even ring transfers.  This way when we design new components we can follow those rules so we don't have to redesign stuff later.  I would be happy to help with this, but I cannot take the lead on it right now as I would like to finish the partial replication component, but before I do that I am trying to get Antidote to work during normal execution.

marc-shapiro:
OK, point taken.

	Marc

tcrain:
I created an issue here https://github.com/SyncFree/antidote/issues/193 to keep track of the fault tolerance discussion.

PR #194, opened 2016-01-21
Support machines with multiple network interfaces

tcrain:
This is to support setups where machines have multiple network interfaces with multiple ips.  The descriptor for a DC now contains the ips for all the interfaces for all the nodes in that DC.  Then when a cross DC connection is created using zero mq it tries each interface is tried until it finds one that succeeds.  If the DCs are unable to establish a connection after a timeout an error is returned.  Since we are using zero mq which abstracts the connection layer we need to send a message manually and wait for an ack.

[This fixes the issue with Grid5000 where we could only connect within a single cluster, as certain machines have several addresses, some of which are valid public internet address, but only accept traffic from within the cluster]

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -128,6 +128,7 @@
 -type dcid() :: term().
 -type tx() :: #transaction{}.
 -type cache_id() :: ets:tid().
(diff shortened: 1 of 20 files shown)


bieniusa:
Can you have a look at the merge conflicts?

tcrain:
I've merged this pull into https://github.com/SyncFree/antidote/pull/195 since they were relying on each other, so I will close this one.

PR #195, opened 2016-01-29
Keep vnodes idle at startup

tcrain:
The idea of this pull request is keep vnodes as light as possible until the ring is joined and to be sure that when the ring is joined they can safely startup any background processes and shared tables.

When the ring is stable rpc:call(Node1, inter_dc_manager, start_bg_processes, [stable]) can be called to any node in the DC to start the necessary background processes.

The reasoning for this is that there seems to be a race condition when too many nodes try to join the ring at the same time as described here
https://github.com/basho/riak_core/issues/405

So the idea of this pull request is to make sure we have nothing that could promote the occurrence of this.  Then as suggested in (https://github.com/basho/riak_core/issues/405), if we add sleeps in between when nodes try to join the ring the ring joins more consistently.  With the following change to riak test things seem to work frequently:
https://github.com/tcrain/riak_test/commit/66af564fc645b364d4087f4f7d1f08fa96d0fef2

I have also tested adding 10 second sleeps in between joining nodes to the cluster on Grid5000 and have been able to consistently join clusters of 40 nodes whereas before I was unable to join 10 nodes frequently.

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -128,6 +128,7 @@
 -type dcid() :: term().
 -type tx() :: #transaction{}.
 -type cache_id() :: ets:tid().
+-type inter_dc_conn_err() :: {error, {partition_num_mismatch, non_neg_integer(), non_neg_integer()} | {error, connection_error}}.
 
 -export_type([key/0, op/0, crdt/0, val/0, reason/0, preflist/0,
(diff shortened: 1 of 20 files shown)


tcrain:
This pull request now also includes the changes described in https://github.com/SyncFree/antidote/pull/194 since they rely on each other.

PR #201, opened 2016-05-02
fix interdc blocking

tcrain:
This small change fixes a bug that could cause the interDC layer to block, resulting in updates from other DCs to buffer forever.  It only happens with 3 or more DCs, and usually needs a high update rate at multiple DCs.  The problem happens when two updates from two different DCs arrive at a third DC and each of them depend on a timestamp for the other DC that is more recent than the latest timestamp received at the third DC.  (Note that the reason why they are not processed until their dependencies are satisfied is so that updates are sent to the materialiser in causal order)

To fix this, when an update cannot be applied due to its dependencies not being met, the "received time" for the DC that sent this update is increased to 1 minus the commit time of this update.  This is safe to do because updates from an external DC are processed in the order of their commit time.

Note that these are per "shard" dependencies, so the visibility of these updates is still governed by the global stable time.

### DIFF
--- a/riak_test/multiple_dcs_test.erl
+++ b/riak_test/multiple_dcs_test.erl
@@ -4,6 +4,7 @@
          multiple_writes/4,
          simple_replication_test/3,
          failure_test/3,
+	 blocking_test/3,
          parallel_writes_test/3]).
 
 -include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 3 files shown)


bieniusa:
Would it be possible to write a (riak-)test to cover the scenario?

tcrain:
I will try to come up with something.  It's a bit tricky because it depends on the timing of messages and physical clocks, but I can probably instrument something.

On this topic though, we should really have some automated test that runs a cluster of several nodes at high load for a few minutes, then at the ends check that the results are 'correct'.  Something like this would be extremely helpful on many levels of debugging, performance testing.

tcrain:
Test added, this blocks in master, but runs here.

Issue #193, opened 2016-01-18
Antidote fault tolerance and recovery


tcrain:
Currently Antidote's primary fault tolerance is through updates being replicated at multiple DCs and that lost messages across DCs are detected and resent.

The design does not ensure safety and progress and recovery when:
A process crashes within a DC, whether that be a vnode process, a maintenance process, or a transaction coordinator.
Messages within the DC are lost.
Any type of ring handoff that happens outside of system initialisation when no transactions are happening.
(Please add more as you notice them)

We should come up with an overall design and some guidelines for ensuring fault tolerance that fits well with the supervisor model.

Here is a comment from Marc about the transaction prepare and abort messages that we should use when designing the model:

"the fate of a transaction logically depends ONLY on unanimous "prepare".  Once all transaction participants have recorded the prepare message in their log, the transaction IS committed, and all entities in the system must reach this same conclusion no matter what happens.  If there is a state where not all participants have recorded prepare, and one knows for sure that this will not happen in the future (e.g. a participant is known to have crashed) then everyone must conclude "abort".  The abort message is an optimisation, and the termination should take place even if participants crash or messages get lost."

This thread can be used as point to keep track of where we encounter fault tolerance problems and the discussion fixing them.

viktoriafordos:
We have been testing Antidote with multiple DCs and have found that after a node failure the replication didn't start again. My guess is the issue is somewhat related to the pub-sub model. For instance, a suspicious TODO comment is [here](https://github.com/SyncFree/antidote/blob/master/src/inter_dc_sub.erl#L53).

Eipifi:
I can help you with this one. When the node starts listening to a remote
DC, there are two mechanisms that take place.
First, the node creates a new SUB socket and connects it to the PUB socket
on the remote DC.
Second, it also creates a REQ socket and connects it to the REP socket on
the remote DC. This REQ/REP socket pair is for fetching all transactions
lost in case of a pub/sub failure.

Now if the node crashes, it needs to reconnect to the exact same PUB
sockets and same REP sockets. The internals will take care of the lost
messages as soon as the newly recovered node receives a ping. It will then
figure out which transactions were lost, redownload them via the REQ/REP
and resume the normal communication via PUB/SUB.

To solve the problem, all that is needed is a persistent storage method for
the DCID -> IP/port mapping.

śr., 9.03.2016 o 10:48 użytkownik Viktória Fördős <notifications@github.com>
napisał:

[...]

PR #202, opened 2016-05-09
Recover and reconnect on restart

tcrain:
The previous pull request (https://github.com/SyncFree/antidote/pull/200) ensured that when an Antidote node failed and restarted it would load updates stored in the log into memory so that they can be read. 

This pull request continues on the recovery theme.  What is added here is that on restart, all previous connections and meta-data states are reestablished.  This includes the interDC connections, and the meta-data that is used to track the operation numbers so that the failed node can both send and receive correctly the messages that might have been lost during the failure.  This data is both loaded from the operation log and a dets file that keeps information about connections.
This pull request is resolving some of the issues described here https://github.com/SyncFree/antidote/issues/193

The recovery can be disabled by setting recover_meta_data_on_start to false in antidote.app.src https://github.com/SyncFree/antidote/blob/stable-meta-data/src/antidote.app.src#L15

New tests are added to test the basic functionality of the recovery.


riak_test/common.erl
```diff
@@ -21,30 +21,42 @@
...
     end.
 
+just_clean_clusters(Clusters) ->
+    disconnect_dcs(Clusters),
+    lists:foldl(fun(Cluster, Acc) ->
+			rt:clean_cluster(Cluster),
+			Acc ++ [length(Cluster)]
+		end, [], Clusters),
+    ok.
+
```
bieniusa: The lists:foldl should be simplified to:

lists:map(fun(Cluster) -> rt:clean_cluster(Cluster))

You never use the accumulated value.

tcrain: Ok, this is fixed.

riak_test/common.erl
```diff
@@ -21,30 +21,42 @@
...
   clean_clusters/1,
+  clean_clusters/2,
+  just_clean_clusters/1,
   setup_dc_manager/2]).
 
 %% It cleans the cluster and formes a new one with the same characteristic of the input one.
 %% In case the clean_cluster parameter is set to false in the riak_test configuration file
 %% the function simply returns the inputed set of clusters
 %% Clusters: A list of clusters as defined by rt
-clean_clusters(Clusters)->
```
bieniusa: I would suggest some renaming: 
clean_and_rebuild_clusters?

riak_test/multiple_dcs_node_failure_test.erl
```diff
@@ -0,0 +1,350 @@
...
+
+    %% Kill and restart a node an be sure everyhing works
+    lager:info("Killing and restarting node ~w", [Node1]),
+    timer:sleep(5000),
+    rt:brutal_kill(Node1),
+    rt:start_and_wait(Node1),
+    rt:wait_until_ring_converged(Cluster1),
+    lager:info("Waiting until vnodes are restarted"),
+    rt:wait_until(Node1, fun wait_init:check_ready/1),
+    timer:sleep(10000),
```
bieniusa: Are these sleeps really necessary? They are rather long.

src/wait_init.erl
```diff
@@ -30,6 +30,7 @@
 %% and gen_servers serving read have been started.
 %% Returns true if all vnodes are initialized for all phyisical nodes,
 %% false otherwise
+-spec wait_ready_nodes([node()]) -> true.
 wait_ready_nodes([]) ->
```
bieniusa: Rewrite using lists:all  ?

src/wait_init.erl
```diff
@@ -42,6 +43,7 @@ wait_ready_nodes([Node|Rest]) ->
 
 %% @doc This calls the check_ready function repatabliy
 %% until it returns true
+-spec wait_ready(node()) -> true.
 wait_ready(Node) ->
```
bieniusa: Rewrite using lists:all() ?

riak_test/multiple_dcs_node_failure_test.erl
tcrain: It's to be sure the updates have been stored to log durably.
Right now the disk log sync option is hardcoded in antidote.hrl, so maybe I should put that as one of the config options here: https://github.com/SyncFree/antidote/blob/master/src/antidote.app.src#L15
so then I can have the sync log be enabled for this test?

PR #205, opened 2016-06-22
Commit hooks

deepthidevaki:
- Users can register pre- and post- commit hooks per bucket. 
- Pre-commit hook will be executed before executing each update 
- Post- commit hooks will be executed after the transaction successfully commits.
- If pre-commit hook failed transaction will be aborted
- If post-commit hook fails, it is ignored. Transaction have already committed.
- commit hooks are executed only in the origin.


include/antidote.hrl
```diff
@@ -130,10 +130,24 @@
 -type cache_id() :: ets:tid().
 -type inter_dc_conn_err() :: {error, {partition_num_mismatch, non_neg_integer(), non_neg_integer()} | {error, connection_error}}.
 
+
```
bieniusa: Can you meanwhile resolve those todos?

src/antidote_hooks.erl
```diff
@@ -0,0 +1,119 @@
...
+        {Module, Function} ->
+            try Module:Function({{Key, Bucket}, Type, Param}) of
+                {ok, Res} -> Res
+            catch
+                _:Reason -> {error, Reason}
+            end
+    end;
+%% The following is kept to be backward compatible with the old
+%% interface where buckets are not used
+execute_pre_commit_hook(Key, Type, Param) ->
```
bieniusa: Spec is missing.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -584,6 +606,17 @@ reply_to_client(SD = #tx_coord_state{from = From, transaction = Transaction, rea
                                 {ok, {TxId, lists:reverse(ReadSet), Transaction#transaction.vec_snapshot_time}}
                         end;
                     committed ->
+                        %% Execute post_commit_hooks
```
bieniusa: Clean-up?

bieniusa:
Is there anywhere a bit of documentation how hooks can be specified?

PR #207, opened 2016-07-11
Async read requests and general code/type/record cleanup

tcrain:
The idea of this branch is to allow for DCs to perform requests different types of requests from one to another in a scalable manner.
For example these requests could include asking for:
- missing log entries
- permissions for needed for specific objects (like bounded counter)
- cross dc read requests during partial replication
- cross dc object validation/certification

Currently only the missing log entries request is implemented.
Note that that log entry requests already existed in Antidote from the ZMQ InterDC implementation, but they were hardcoded as the only type of requests and the were sequential and blocking per DC.
Now different types of requests can be added by modifying:
antidote_message_types.hrl, inter_dc_query.erl, and inter_dc_query_receive_socket.erl
Requests and replies are done asynchronously and handled by a pool of threads and are sent directly to the node responsible for the key/partition of the request.

This pull request also includes lots of code cleanup and commenting and dialyzer additions (when I was making the log requests parallel I noticed these issues so I fixed a lot of them at the same time).
This includes better defined types and records thought the code, especially in the log.  This lead to a lot of dialyzer fixes as well (we had a lot of missing entries or wrong entries that were still passing because many were defined incorrectly, but there is still more to be done here).  Note that we are using a version of Erlang from 3 years ago, so we is seems we cannot use some of the new ways of typing objects (for example we cannot type dicts), I still added some of these new definitions (but in comments) in the logging code as I think they help make the code a lot more understandable.

Additionally I moved some of the code to a separate project called antidote_utils (https://github.com/SyncFree/antidote_utils).  The reason for this is that as were are breaking Antidote into separate projects we will need to share this code (for example for the the separate logging project).  It currently includes vector clocks, vector clock ordered dictionaries, and some type definitions.


bieniusa:
Thanks for all this work - but PLEASE next time delta merge requests. This is just too much stuff to review ...

Issue #209, opened 2016-08-04
Lost updates


peterzeller:
Updates can be lost in the materializer as shown by the following riak_test case: https://gist.github.com/peterzeller/1e5a6ad396bbd9b6831072a94f49e65e
The test is not deterministic, but it will often miss number 10017, 10018, or 10019, which is directly after the garbage collector runs for the first time.

The problem seems to be in the assumption that vector_orddict is an orddict, but it is not.

For example, this assumption is used in [clocksi_materializer, line 160](https://github.com/SyncFree/antidote/blob/44edb57e8e63e3d8e72794bc84c93ce894c5a39f/src/clocksi_materializer.erl#L160), but it does not hold.

The order is reversed in several places, for example in the garbage collector in [reverse_and_filter](https://github.com/SyncFree/antidote/blob/eaf5e7e0e636d26991539723d7764b3f5a0d2442/src/materializer_vnode.erl#L486).

Some of the problems were maybe introduced with pull request https://github.com/SyncFree/antidote/pull/185

tcrain:
Thanks for making the test, it helped to find what was going on.

It looks like the bug was that after garbage collection the list of remaining operations was left in the reverse order than what it should be.  There is a function in the materializer "reverse_and_filter" https://github.com/SyncFree/antidote/blob/master/src/materializer_vnode.erl#L518 that was actually "incorrectly" reversing the list of operations.  It looks like I introduced this bug when we switched from using lists to tuples in the materializer which required using a different way to construct the tuple.

I added the fix plus (a slightly modified version of) your test here, see if it works for you:
https://github.com/SyncFree/antidote/commit/28d463badb58e1e4475ec4058aae4af992e76ba6

The reason that none of our current tests caught the bug was that they all used only the most recent snapshot after garbage collection, so it they didn't need to apply any of the old ops after gc.  We should probably make a "concurrency/load suite" of tests to catch things like this.

I will also try to clean up the materializer code a bit and add some text explaining how it works.

tcrain:
Also, just for info, the reason why lists are being reversed is here:
http://erlang.org/doc/efficiency_guide/listHandling.html#id68235

peterzeller:
It works for me now. Thanks for the fast fix.

tcrain:
Fixed in https://github.com/SyncFree/antidote/pull/210

PR #210, opened 2016-08-26
Fix list reverse materializer

tcrain:
This fixes the bug described in https://github.com/SyncFree/antidote/issues/209

The bug was that the list of operations cached in the materialiser after garbage collection was stored incorrectly in reverse order.  The test case from that issue is included here as well.

To explain further, updates are stored in the materialiser cache in a tuple.  The garbage collection was converting this tuple into a list and then filtering out old operations.  The fix in this pull requests performing the filtering by going directly through the tuple, avoiding an extra tuple to list conversion and list reversal.

This pull request also makes the following other small changes to the materialiser:

The materialiser now avoids converting the cached tuple to a list during a read operation (so avoiding another unnecessary conversion).

An extra check is added when inserting new snapshots in the cache, so that only new snapshots are inserted in the cache. (Previously two concurrent read operations could insert the same snapshot twice)


tcrain:
This is now up to date with the rebar 3 changes.

The multiple_dcs_SUITE test is added and all seem to be passing.

bieniusa:
Now it all passes :)

PR #220, opened 2016-09-27
[WIP] Erlang 19 integration branch

cmeiklejohn:


include/antidote.hrl
```diff
@@ -111,14 +111,17 @@
...
+  global :: undefined | non_neg_integer(),
+  local :: undefined | non_neg_integer()}).
 
 %% The way records are stored in the log.
 -record(log_record,{
 	  version :: non_neg_integer(), %% The version of the log record, for backwards compatability
-	  op_number :: #op_number{},
-	  bucket_op_number :: #op_number{},
-	  log_operation :: #log_operation{}}).
+	  op_number :: undefined | #op_number{},
```
cmeiklejohn: Why precisely would these ever be undefined?  How about instead of modifying the type signature, we identify the location in the code where they are being initialized to undefined and change that to use a starting value.

What do you think?

include/antidote.hrl
```diff
@@ -111,14 +111,17 @@
...
+	  bucket_op_number :: undefined | #op_number{},
+	  log_operation :: undefined | #log_operation{}}).
```
cmeiklejohn: Why precisely would these ever be undefined? How about instead of modifying the type signature, we identify the location in the code where they are being initialized to undefined and change that to use a starting value.

What do you think?

include/antidote.hrl
```diff
@@ -111,14 +111,17 @@
...
+  local :: undefined | non_neg_integer()}).
 
 %% The way records are stored in the log.
 -record(log_record,{
 	  version :: non_neg_integer(), %% The version of the log record, for backwards compatability
-	  op_number :: #op_number{},
-	  bucket_op_number :: #op_number{},
-	  log_operation :: #log_operation{}}).
+	  op_number :: undefined | #op_number{},
+	  bucket_op_number :: undefined | #op_number{},
```
cmeiklejohn: Why precisely would these ever be undefined? How about instead of modifying the type signature, we identify the location in the code where they are being initialized to undefined and change that to use a starting value.

What do you think?

include/antidote.hrl
```diff
@@ -230,6 +233,6 @@
 %% clocksi_read_item_fsm as pointers to the materializer's ets caches
 -record(mat_state, {
 	  partition :: partition_id(),
-	  ops_cache :: cache_id(),
-	  snapshot_cache :: cache_id(),
-	  is_ready :: boolean()}).
+	  ops_cache :: atom() | cache_id(),
```
cmeiklejohn: Can we be more specific than `atom()` here?

rebar.config
```diff
@@ -1,24 +1,35 @@
...
+    {antidote_pb, {git, "git://github.com/syncfree/antidote_pb", {branch, "new_interface_erl19"}}},
+    {antidote_crdt, ".*", {git, "git://github.com/syncfree/antidote_crdt", {tag, "0.0.2"}}},
     {antidote_utils, {git, "git://github.com/syncfree/antidote_utils", {tag, "0.0.1"}}}
 ]}.
 
 {erl_opts, [debug_info, warnings_as_errors, {parse_transform, lager_transform},{i,"include"}]}.
 {cover_enabled, true}.
 {eunit_opts, [verbose, {report, {eunit_surefire, [{dir,"logs/"}]}}]}.
+{dialyzer, [{warnings, [
+  %unmatched_returns,
```
cmeiklejohn: Why?

src/clocksi_interactive_tx_coord_sup.erl
```diff
@@ -34,8 +34,8 @@ start_link() ->
     supervisor:start_link({local, ?MODULE}, ?MODULE, []).
 
 start_fsm(Args) ->
-    _Res = random:seed(dc_utilities:now()),
-    Random = random:uniform(?NUM_SUP),
+    _Res = rand:seed(exsplus, erlang:timestamp()),
```
cmeiklejohn: This is a backwards incompatible change, right?  For instance, will this compile with Erlang 18?  (That's the most commonly deployed Erlang.)

include/antidote.hrl
peterzeller: That would probably be the better solution, but just changing the type has the benefit that it cannot break any functionality. Adding undefined here is just emulating what previous versions of Erlang did automatically for record-fields without default value.

include/antidote.hrl
peterzeller: In `clocksi_readitem_fsm` this field is initialized with a value from `materializer_vnode:get_cache_name`, which constructs an atom dynamically. So if there is a type that says the atom ends with `"snapshot_cache"`, then we could make it more precise. 

I think the `cache_id()` alternative could be removed.

rebar.config
peterzeller: I had to add the `no_undefined_callbacks` option, because there is some dialyzer error in riak_dt, which leads to an error "Callback info about the riak_dt behaviour is not available".

When I added that option, I looked at the other options in the dialyzer manual and added some (commented) options which might be useful to add later. However, I did not want to mix Erlang19 changes with other typing stuff.

peterzeller:
> I'm confused why you're not using rand_compat here? I think that we want to support both Erlang 18 and Erlang 19, generally.

The discussion in Paris went back and forth a few times, but as I understood it, the final decision was to not spend any time on compatibility with older versions until we have a first stable release. And after the Antidote 1.0 release we would support the most commonly used versions.

Changing it to your compatibility library would not be much effort, so I could do that.

peterzeller:
What still needs to be discussed is the time handling. 

Here is what I got now from the discussions and from looking at the code.

- Monotonicity: Replacing `erlang:now` with `erlang:system_time` (as it is done currently) should work, as `system_time` is still monotonic with the default time warp setting ("No Time Warp Mode").
- Uniqueness: Most old uses of `erlang:now` do not need uniqueness. I think it might still be required for generating transaction-ids (in `clocksi_interactive_tx_coord_fsm`, line 171). There might be other places.
- Restarts:  `erlang:now` and `erlang:system_time` seem to loose their guarantees after a system restart. So I guess the current version is also not correct in that sense. 

Overall, I would suggest to fix the time handling in a separate pull request, since it probably requires a few refactorings if we really want to make the code safe in the presence of time warps and crashes.

include/antidote.hrl
bieniusa: I agree with @cmeiklejohn to fix this now. Otherwise, it might create odd behavior eventually. Let's stay safe here.

include/antidote.hrl
bieniusa: Let's file an issue for that. Dynamic creation of atoms should be considered harmful and at least inspected with care.

rebar.config
bieniusa: Ok, let's add this for later todos.

src/clocksi_interactive_tx_coord_sup.erl
bieniusa: Can we use your compatibility library here?

bieniusa:
I agree with the separate pull request. We need to address fault-tolerance in detail, and separation of concern is here the best way to make progress.

peterzeller:
I changed the uses of `rand` to `rand_compat`.

Regarding the `undefined`s in the record types: There are  several cases:

1. I have changed the cases, where only minimal code changes were required. 
2. I have added comments starting with `TODO 19` for the cases that require bigger code changes, or more  knowledge about the purpose of the code.
3. Some fields are really optional, like the `commit_time` in the `tx_coord_state`.
4. One field was not used at all.

cmeiklejohn:
Is this ready to be merged?  What work still remains?  If any, can we make a check list in this PR so we know what we're waiting on?

peterzeller:
If I understood @bieniusa correctly, she would also handle the erlang-time problems separately. Then I think this is ready to be merged.

cmeiklejohn:
@bieniusa can you confirm?

bieniusa:
Confirm hereby the mergability.

@Peter: Can u take care that the remaining issues and bugs are tracked?

Von meinem iPhone gesendet

aletomsic:
Do we need to remove our old erlang and get 19 now?
Is the change back compat to 16B02?

peterzeller:
I tried it with Erlang 18.3 and 19.0. It is not compatible with Erlang 16.

aletomsic:
Problem: after the change, this happens when running antidote console:

[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
{"Kernel pid terminated",application_controller,"{application_start_failure,riak_core,{bad_return,{{riak_core_app,start,[normal,[]]},{'EXIT',{undef,[{cuttlefish_schema,files,[[\"/Users/alek/test_antidote/antidote/_build/default/rel/antidote/lib/11-erlang_vm.schema\",\"/Users/alek/test_antidote/antidote/_build/default/rel/antidote/lib/12-riak_core.schema\",\"/Users/alek/test_antidote/antidote/_build/default/rel/antidote/lib/15-riak_sysmon.schema\",\"/Users/alek/test_antidote/antidote/_build/default/rel/antidote/lib/21-leveldb.schema\"]],[]},{clique_config,load_schema,1,[{file,\"/Users/alek/test_antidote/antidote/_build/default/lib/clique/src/clique_config.erl\"},{line,81}]},{riak_core_cli_registry,load_schema,0,[{file,\"/Users/alek/test_antidote/antidote/_build/default/lib/riak_core/src/riak_core_cli_registry.erl\"},{line,53}]},{riak_core_app,start,2,[{file,\"/Users/alek/test_antidote/antidote/_build/default/lib/riak_core/src/riak_core_app.erl\"},{line,104}]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,273}]}]}}}}}"}
Kernel pid terminated (application_controller) ({application_start_failure,riak_core,{bad_return,{{riak_core_app,start,[normal,[]]},{'EXIT',{undef,[{cuttlefish_schema,files,[["/Users/alek/test_antidot

Crash dump is being written to: erl_crash.dump…done

aletomsic:
Using erlang 19.

bieniusa:
Which Erlang exactly are you using? 
I had problems with the one from ESL (was missing a Library), but can run tests with the one from Ericsson.

Can u compile and run the systests?

Von meinem iPhone gesendet

> Am 07.10.2016 um 13:40 schrieb Alejandro Zlatko Tomsic <notifications@github.com>:
> 
> Using erlang 19.
>

aletomsic:
> On Oct 7, 2016, at 13:45, Annette Bieniusa <notifications@github.com> wrote:
> 
> Which Erlang exactly are you using? 

19.1 rev3 from ESL. The latest available.

> I had problems with the one from ESL (was missing a Library), but can run tests with the one from Ericsson.

I have just tried uninstalling, installing 19 from ESL. Does not work.
I uninstalled that, installed 19 from macports, as suggested in http://www.erlang.org/downloads <http://www.erlang.org/downloads>.

Same problem.

> 
> Can u compile and run the systests?
> 
> Von meinem iPhone gesendet
> 
> > Am 07.10.2016 um 13:40 schrieb Alejandro Zlatko Tomsic <notifications@github.com>:
> > 
> > Using erlang 19.
> >

aletomsic:
> On Oct 7, 2016, at 14:09, Alejandro Z. Tomsic <aletomsic@gmail.com> wrote:
> 
> 
>> On Oct 7, 2016, at 13:45, Annette Bieniusa <notifications@github.com <mailto:notifications@github.com>> wrote:
>> 
>> Which Erlang exactly are you using? 
> 
> 19.1 rev3 from ESL. The latest available.
> 
>> I had problems with the one from ESL (was missing a Library), but can run tests with the one from Ericsson.
> 
> I have just tried uninstalling, installing 19 from ESL. Does not work.
> I uninstalled that, installed 19 from macports, as suggested in http://www.erlang.org/downloads <http://www.erlang.org/downloads>.
> 
> Same problem.
> 
>> 
>> Can u compile and run the systests?

yes, can compile and all tests pass.

>> 
>> Von meinem iPhone gesendet
>> 
>> > Am 07.10.2016 um 13:40 schrieb Alejandro Zlatko Tomsic <notifications@github.com <mailto:notifications@github.com>>:
>> > 
>> > Using erlang 19.
>> >

deepthidevaki:
I am also getting similar error when starting antidote rel. Tests are
running fine.
I have erlang 19.1 from erlang.org.

There was a similar issue when moved to rebar3. These applications had to
be explicitly added to app.src for them to be loaded.

peterzeller:
If it is a problem with the started application, we should maybe revert my commit here: https://github.com/SyncFree/antidote/pull/220/commits/944c7ff83dbad8167c8add55600ffd6877050c65

Could we extend the system tests to detect problems like this?

bieniusa:
Actually, we fixed it in so far that we can get a console running by using the release:

./_build/default/rel/antidote/bin/antidote console

However, running „rebar3 shell“ does not work, even though the documentation says that it uses a release if there is one.

Issue #229, opened 2016-10-10
Update rebar.config to include latest version of antidote_crdt


aletomsic:
Hi, I've added the antidote_crdt_lwwreg to the list of CRDTs in: 

https://github.com/SyncFree/antidote_crdt/blob/master/src/antidote_crdt.erl#L35

Without that, the type check fails when issuing an operation to that type, as it believes the type does not exist.

What we are missing now is that Antidote's rebar.config points to the latest changes of antidote_crdt. 

Should we build a new release of antidote_crdt? 
So far, I am using, in the PhysiCS branch, this:
https://github.com/SyncFree/antidote/blob/PhysiCS/rebar.config#L12

instead of what master uses: 
https://github.com/SyncFree/antidote/blob/master/rebar.config#L12

bieniusa:
We should probably do this once @peterzeller has updated the antidote_crdt repo.

peterzeller:
The crdt library was updated with #230

PR #230, opened 2016-10-10
Update CRDT-lib and protocol buffer interface

peterzeller:
* updated CRDT library (also includes #229)
* updated protocol buffer library
* clock in start_transaction is now optional, omitting the value is equivalent to `ignore` (this required a few changes in `antidote_pb_txn.erl`)
* some new protocol buffer system tests
* some protocol buffer system tests were changed: the new pb-client version only allows to store binary values in sets/registers


bieniusa:
@peterzeller If you don't need this branch anymore, please delete!

PR #234, opened 2016-10-13
[WIP] Deployment Branch

cmeiklejohn:
- [x] Docker image for deployment.
- [ ] Script to launch multiple clusters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ docs/_pdf
 docs/.idea/
 log/
 data/
+compile_commands.json
--- a/Dockerfiles/antidote
+++ b/Dockerfiles/antidote
@@ -1,0 +1,19 @@
+FROM ubuntu:xenial
+
+MAINTAINER Christopher Meiklejohn <christopher.meiklejohn@gmail.com>
+
+RUN cd /tmp && \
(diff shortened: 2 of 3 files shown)


mweberUKL:
Seems like you start the console for starting Antidote. This means that you can only start in interactive mode.

Can you maybe reuse some parts of the script I wrote for starting an Antidote-node standalone in docker? This should also resolve the problem of connecting to the node using RPC (if needed) because it fixes the RPC port to be 9100.

See my repo: https://github.com/mweberUKL/docker-antidote

cmeiklejohn:
We're going to use the code we have for Lasp that automates all of this for Amazon.

peterzeller:
Maybe we should also add Matias' Dockerfile to the repo or to the documentation. For end-users it seems to be better, since it is smaller, sets up the ports correctly, can be started in noninteractive mode, and has some variables to configure the node names etc.

cmeiklejohn:
Yes, I suggest that his Dockerfile is added into the `Dockerfiles` directory with the name `antidote-local` so it's clear that it can be used to run a local node.

cmeiklejohn:
Abandoning.

PR #235, opened 2016-10-16
[WIP] Cleanup, and release integration.

cmeiklejohn:
- [x] Rebar3 and relx are not being used properly to build the release.
- [x] Ports do not have defaults and cause crashes if the ports are omitted.
- [x] Docker container for Mesos deployment that uses dynamic port configuration.
- [x] `make stage` doesn't use overlay configuration, but release does.
- [x] `rebar3 shell` will be resolved once proper relx configuration is done.
- [x] Use built-in `foreground` command from relx for launching the container instances, instead of a custom script.
- [x] Fork `riak_core_ng` and build a version without the `erocksdb` dependencies.


cmeiklejohn:
@bieniusa Please review.

cmeiklejohn:
At this point, there's no way to fix the `rebar shell` because of requirements coming from cuttlefish.

PR #237, opened 2016-10-19
doku: instructions use new erlang versions

deepthidevaki:


bieniusa:
Maybe update also the mailing list to antidote-support?

PR #242, opened 2016-10-23
Launch multiple nodes.

cmeiklejohn:
It's no longer necessary to build multiple releases to run multiple
instances of the system, as you can use the dynamic configuration as
supported by rebar3.


bieniusa:
Not happy that Travis is showing so many issues here, but I don't think they are related.

PR #243, opened 2016-10-24
Add console target.

cmeiklejohn:
Add a target that launches the console using the dynamic configuration provided by relx.


bieniusa:
We need to update the documentation accordingly!

bieniusa:
Also need to adapt the output of log messages.
I think we already did a pass to check that the code uses appropriate logging levels, but we need to configure lager to only show warning and worse for the console at least.

This will also be an issue for the benchmarking. Too much logging introduces noticeable overhead.

cmeiklejohn:
You should be *removing* unnecessary logging statements: even if something is at a lower level than what should be displayed, you still take the performance hit of dispatching the messages and ignoring it.  This was a huge issue in Riak.

cmeiklejohn:
Also, I'd really appreciate someone else picking up the documentation changes, if possible.

peterzeller:
Can't we use the same script for `env` and `ienv` and just add a parameter for foreground/start/console in the last line? The files are nearly identical except for the last line and two more options in ienv.

> You should be removing unnecessary logging statements: even if something is at a lower level than what should be displayed, you still take the performance hit of dispatching the messages and ignoring it. This was a huge issue in Riak.

Doesn't lager have an option to do this filtering with a parse-transform at compiletime?

cmeiklejohn:
> Can't we use the same script for env and ienv and just add a parameter for foreground/start/console in the last line? The files are nearly identical except for the last line and two more options in ienv.

Sure, can you make the change?  I won't for another few days.

> Doesn't lager have an option to do this filtering with a parse-transform at compiletime?

It didn't last time I looked, but I haven't looked at lager since at least a major revision ago.

cmeiklejohn:
I cleaned up the `env` and `ienv` and I'm going to merge for now so we can get this ready for the tutorial, and will fix the documentation afterwards.

deepthidevaki:
`make console` gives me a segmentation fault.

./_build/default/rel/antidote/bin/env console
IP address not set; defaulting to 127.0.0.1.
MESOS_TASK_ID: 
PORT0: 
PORT1: 
PORT2: 
PORT3: 
PORT4: 
NODE_NAME: antidote@127.0.0.1
COOKIE: antidote
IP: 127.0.0.1
HOSTNAME: supernova
Exec: /home/deepthi/work/test/antidote/_build/default/rel/antidote/erts-8.1/bin/erlexec -boot /home/deepthi/work/test/antidote/_build/default/rel/antidote/releases/0.0.1/antidote -mode embedded -boot_var ERTS_LIB_DIR /home/deepthi/work/test/antidote/_build/default/rel/antidote/erts-8.1/../lib -config /home/deepthi/work/test/antidote/_build/default/rel/antidote/releases/0.0.1/sys.config -args_file /home/deepthi/work/test/antidote/_build/default/rel/antidote/releases/0.0.1/vm.args -- console
Root: /home/deepthi/work/test/antidote/_build/default/rel/antidote
/home/deepthi/work/test/antidote/_build/default/rel/antidote
Erlang/OTP 19 [erts-8.1] [source] [64-bit] [smp:4:4] [async-threads:5] [hipe] [kernel-poll:true]

make: *** [console] Segmentation fault (core dumped)

cmeiklejohn:
All `make console` is doing is calling console after setting up the environment.  Can you try to debug it further, there's no way to debug given the information you've provided.

PR #247, opened 2016-11-01
Makefile target for rebar3 shell

peterzeller:
The `rebar3 shell` was no longer working.

I created a special config file `sys-debug.config` for use with the shell. It is similar to the `sys.config` used for releases, but does not use environment variables. I also had to add the `schema_dirs` configuration for riak_core to make it work.

Besides the `make shell` target, I also added a `make auto` target for a shell which automatically compiles and reloads the code when it is changed on disk (see https://github.com/vans163/rebar3_auto).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ docs/.idea/
 log/
 data/
 compile_commands.json
+data.*/
+log.*/
+erl_crash.dump
+rebar3.crashdump
+test/.rebar3/erlcinfo
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,14 @@ cleantests:
 	rm -rf logs/
 
 shell:
-	$(REBAR) shell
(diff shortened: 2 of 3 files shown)


peterzeller:
Got two errors when trying on other machines:

## 1. 
```
 ===> Failed to boot riak_core for reason {bad_return,
                                                     {{riak_core_app,start,
                                                       [normal,[]]},
                                                      {'EXIT',
                                                       {{function_clause,
                                                         [{orddict,fetch,
                                                           ['antidote@127.0.0.1',
                                                            [{nonode@nohost,
                                                              [{{riak_core,
                                                                 bucket_types},
                                                                [true,false]}, 
```

Deleting the `data*` folders from the antidote directory fixed that problem.


## 2. 

> Erlang Distribution failed, falling back to nonode@nohost. Verify that epmd is running and try again.


Running `epmd &` before `make shell` fixed this.

----

I am not sure if there is a nice way to handle these problems in the Makefile.

cmeiklejohn:
:+1:, thanks @peterzeller

Issue #248, opened 2016-11-05
Operations that affect multiple items done sequentially


aletomsic:
The new interface: (start_transaction, read_objects, update_objects) calls the interactive transaction coordinator to issue operations. 
These operations are run today sequentially. E.g., If I issue a read_objects that reads 100 objects, it will send each request to responsive vnodes one by one, after receiving the previous response.
This can be done in parallel, which is done today only at the static_coordinator, for reads.

Furthermore, the static_tx_coordinator is only called by the antidote:execute_tx function, which is deprecated.

What I propose is, 

- handling reads and updates in parallel at the interactive coordinator.
- Make the execute_tx call the interactive coordinator (for backwards compatibility with tests).
- remove the static coordinator, which has tons of duplicate code and logic with the interactive one.

I can take care of this.

Issue #232, opened 2016-10-12
Subsequent txn fail to start after an aborted txn in console


deepthidevaki:
When the following code is executed from test, everything works fine. But when it is executed from antidote console, transaction 2 fail to start.
BCounterObj = {my_bcounter, antidote_crdt_bcounter, my_bcbucket}.
{ok, TxId} = antidote:start_transaction(ignore, []).
{error, _ } = antidote:update_objects([{BCounterObj, decrement, {1,actor}}], TxId). %% decrement not allowed
{ok, TxId2} = antidote:start_transaction(ignore, []).
*\* exception error: no match of right hand side value 
                    {error,{#Ref<0.0.3.2681>,
                            {aborted,{tx_id,1476290149161229,<0.1786.0>}}}}

Why does it happen:
When the update fails txn immediately replies with an error message and initiate abort. Once abort is final, it sends a second reply which is received by the following start_transaction.
It is not visible in tests because (probably) two rpc calls are handled by different processes, so there is no process to receive second reply.

PR #250, opened 2016-11-05
Parallel updates

aletomsic:
This partially addresses issue #248.

A separate PR will address reads.

Before, if you issue an update_objects(ListOfKeyValuePairs), the updates in the list where executed one by one: 
- for each update in the list, 
     - send a read message to the vnode to generate the downstream op (regardless of the type of object and regardless of the object being read before by the transaction), 
    - get the response, generate the downstreamop (at the coordinator), 
    - send a sync message to the vnode for it to log it. 
    - get the response that the logging was ok.

Now: 
- for each update in the list, 
    - (this check is addressed by a separate PR) if the object was not read before (when benchmarking is important to avoid this)
        - send a read message to the vnode to generate the downstream op 
        - get the response, 
    - generate the downstreamop (at the coordinator), 
    - (asynchronously) send a message to the vnode for it to log it. 
- get the ok response of all vnodes.


aletomsic:
By mistake.
Uncommented.

src/logging_vnode.erl
```diff
@@ -346,11 +345,15 @@ handle_command({read_from, LogId, _From}, _Sender,
 %%
 %% -spec handle_command({append, log_id(), #log_operation{}, boolean()}, pid(), #state{}) ->
 %%                      {reply, {ok, #op_number{}} #state{}} | {reply, error(), #state{}}.
-handle_command({append, LogId, LogOperation, Sync}, _Sender,
+handle_command({append, LogId, LogOperation, Sync}, Sender, State)->
+	handle_command({append, LogId, LogOperation, Sync, ignore}, Sender, State);
+
+
+handle_command({append, LogId, LogOperation, Sync, ReplyTo}, _S,
```
deepthidevaki: Isn't ReplyTo same as Sender?

src/logging_vnode.erl
```diff
@@ -384,19 +387,30 @@ handle_command({append, LogId, LogOperation, Sync}, _Sender,
...
+	%% to reply to the process calling that function.
+	%% This is currently being used by the update_objects function in the
+	%% interactive coordinator fsm.
+	case ReplyTo of
+		ignore->
+			{reply, Reply, NewState};
+		_->
+			gen_fsm:send_event(ReplyTo, Reply),
+			{noreply, NewState}
+	end;
```
deepthidevaki: I think this additional ReplyTo is unnecessary. Just {reply, Reply, NewState} and using command/sync_command for async/sync calls gives you the desired behaviour.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -362,18 +345,55 @@ execute_op({OpType, Args}, Sender,
...
+receive_logging_responses(Response, S0 = #tx_coord_state{num_to_read = NumToReply,
+	                        return_accumulator= ReturnAcc}) ->
+	NewAcc = case Response of
+		{error, Reason} -> {error, Reason};
+		{ok, _OpId} -> ReturnAcc;
+		timeout -> ReturnAcc
+	end,
+	case NumToReply > 1 of
+		false ->
+			gen_fsm:reply(S0#tx_coord_state.from, NewAcc),
```
deepthidevaki: Here, is it a reply to the client before a successful commit?

aletomsic: This only replies ok to the client after the update operation was done. 
This does not mean that the transaction has been committed.

src/logging_vnode.erl
aletomsic: No.

src/clocksi_interactive_tx_coord_fsm.erl
deepthidevaki: It should be in "true" of the following case statement. Otherwise if it is an error, it replies a message here and then initiates abort where it sends a reply again.

src/logging_vnode.erl
aletomsic: I'm fixing this as we discussed.

src/antidote.erl
```diff
@@ -126,33 +126,59 @@ read_objects(Objects, TxId) ->
...
-       false -> ok
+    {_, _, CoordFsmPid} = TxId,
+    Operations = lists:map(fun(Update) ->
+                                {Key, Type, Bucket, Op} = case Update of
+                                    %% the following checks the format of the
+                                    %% operation for compatibility with
+                                    %% some systests that call updates with
+                                    %% {Operation, Params} (as a single parameter),
+                                    %% and Operation, Params (two separate parameters).
+                                    {{K, T, B}, O} ->
```
bieniusa: I think fixing the tests would be more sensible..., if not now then in a test refactoring step.
But why has this not been an issue before?

src/antidote.erl
```diff
@@ -126,33 +126,59 @@ read_objects(Objects, TxId) ->
...
+                            end, Updates),
+    case lists:member({error, type_check}, Operations) of
+        true ->
+            {error, type_check};
+        false ->
+            %%            lager:info("gonna start multiple updates: ~p", [Operations]),
+            case gen_fsm:sync_send_event(CoordFsmPid, {update_objects, Operations}, ?OP_TIMEOUT) of
+                ok ->
+                    ok;
+                {aborted, TxId} ->
```
bieniusa: This will cause a type error as the type spec of the function does not expect an {aborted, TxId}, no?

src/antidote.erl
```diff
@@ -126,33 +126,59 @@ read_objects(Objects, TxId) ->
...
+    {ok, vectorclock:new()};
 update_objects(Clock, _Properties, Updates, StayAlive) ->
-    Operations = lists:map(
-                   fun({{Key, Type, Bucket}, Op, OpParam}) ->
-                           {update, {{Key, Bucket}, Type, {Op,OpParam}}}
-                   end,
+    Operations = lists:map(fun(Update) ->
+                                {Key, Type, Bucket, Op} =
+                                    case Update of
+                                    {{K, T, B}, O} ->
```
bieniusa: Same as above.

bieniusa: If it is really, really needed, you should refactor it into a separate function.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -302,36 +304,14 @@ perform_update(Args, Updated_partitions, Transaction, Sender, ClientOps) ->
                                 lists:keyreplace(IndexNode, 1, Updated_partitions,
                                                  {IndexNode, [{Key, Type, DownstreamRecord} | WriteSet]})
                         end,
-                    case Sender of
```
bieniusa: Why can this be deleted? What was the sender used for?

test/clocksi_SUITE.erl
```diff
@@ -130,7 +132,7 @@ clocksi_test1(Config) ->
                       {read, {Key1, Type}}]]),
     ?assertMatch({ok, _}, Result2),
     {ok, {_, ReadSet2, _}}=Result2,
-    ?assertMatch([0,0], ReadSet2),
+    ?assertMatch([0,1], ReadSet2),
```
bieniusa: Why is this change in the test necessary?
We should not fix tests and modify code in the same pull request unless there is a good reason.

src/antidote.erl
aletomsic: yes. I will fix that. Strangely, dialyzer does not complain...
Maybe it can't detect the case where this would happen, as it involves a complex exchange of messages...

src/antidote.erl
aletomsic: Ok, I can create this separate function and when we change the tests remove it.

src/clocksi_interactive_tx_coord_fsm.erl
aletomsic: This was used before to replying to the client here, when issuing single operation updates.
Now the reply is done in the "receive logging responses" state.

test/clocksi_SUITE.erl
aletomsic: For some reason, that test was "wrong". 
It tests that you read your updates within the same transaction. 
Whithin a tx, it: reads an object, updates it, and reads it again, and expects the value to be, wrongly, the same.

The previous path (mentioned in the above discussion), that used execute_tx had a bug that update operations were not applied.

In this case I believe we need to fix the test too.

src/antidote.erl
peterzeller: It would be reported with the `underspecs` option:

> 126: Type specification
> `antidote:update_objects([{bound_object(),op_name(),op_param()}],txid()) -> 'ok' | {'error',reason()}` is a subtype of the success typing: `antidote:update_objects([any()],{_,_,_}) -> 'ok' | {'aborted',{_,_,_}} | {'error',_}`

... but that option also gives a lot of other errors, some of them probably be false positives.

aletomsic:
This PR also addresses issue #232.

src/antidote.erl
```diff
@@ -148,37 +147,68 @@ update_objects(Updates, TxId) ->
...
-                        end;
-                    [_H|_T] -> false
-                end,
-    case SingleKey of
-        true ->  %% if single key, execute the fast path
-            [{update, {K, T, Op}}] = Operations,
-            case append(K, T, Op) of
-                {ok, {_TxId, [], CT}} ->
-                    {ok, CT};
+    case check_and_format_ops(Updates) of
```
aletomsic: This changes make the interactive coordinator execute an update_objects transaction in a static way, similarly to how the static coordinator was doing it, but unified the code in the interactive one, which includes the parallel updates optimization.

src/antidote.erl
```diff
@@ -307,30 +337,14 @@ read(Key, Type) ->
 -spec clocksi_execute_tx(Clock :: snapshot_time(),
                          [client_op()],snapshot_time(),boolean()) -> {ok, {txid(), [snapshot()], snapshot_time()}} | {error, term()}.
 clocksi_execute_tx(Clock, Operations, UpdateClock, KeepAlive) ->
```
aletomsic: This function will be deprecated once I apply the same changes to reads.
This is, making reads parallel, and making read_only_transactions be handled entirely by the interactive txn coord.

src/clocksi_interactive_tx_coord_fsm.erl
```diff
@@ -93,6 +96,15 @@
 %%% API
 %%%===================================================================
```
aletomsic: The following changes add the possibility to start a static transaction that will be entirely handled by this coord.

Issue #249, opened 2016-11-05
Previous reads not used to generate downstream


aletomsic:
The current implementation does not use previous reads done by the transaction to generate downstream operations.
E.g., A transaction reads x, then it updates it. 
The update will send a read to the vnode which is unnecessary because that read has been done.

I propose to have an internal read set at the coordinator for this purpose.

aletomsic:
I see the problem. I was merging the wrong branch. Sorry!

PR #251, opened 2016-11-05
Add readset to generate downstream operations at the coordinator

aletomsic:
This PR addresses issue #249.
Creates an internal read set at the interactive tx coordinator in order to avoid sending read messages to the clocksi_vnode to generate downstream, when the object has already been read by the transaction (which is expected to be frequent in practice).

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -213,6 +213,7 @@
           state :: active | prepared | committing | committed | undefined
                  | aborted | committed_read_only,
           operations :: undefined | list(),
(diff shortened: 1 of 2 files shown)


aletomsic:
Travis is complaining:

===> Compiling apps/antidote/src/vectorclock_vnode.erl failed
include/antidote.hrl:216: type orddict() undefined

It happens here (and does not happen in my machine):
https://github.com/SyncFree/antidote/blob/AddCoordReadSet/include/antidote.hrl#L216

How should I use the orddict as a type?

PR #253, opened 2016-11-06
Fix read order for batch-read

peterzeller:
When reading multiple objects the values are currently returned in reverse order.

I removed two calls to `lists:reverse` to fix this. I am not sure if this is the right place to change it.
Maybe @marsleezm can have a look at it (the code was commited in https://github.com/SyncFree/antidote/commit/6e23cb56cf8be9180858135e54193ebf4ed5bcfc).

@aletomsic I've seen that you changed something with the read-set in #250 
If you already fixed this, you can just copy my testcase to your branch and close this pull-request.

### DIFF
--- a/src/clocksi_static_tx_coord_fsm.erl
+++ b/src/clocksi_static_tx_coord_fsm.erl
@@ -198,11 +198,11 @@ receive_prepared({ok, {Key, Type, Snapshot}},
                     case NumToCommit of
                         0 ->
                             clocksi_interactive_tx_coord_fsm:reply_to_client(S0#tx_coord_state{state=committed_read_only, 
-                            read_set=lists:reverse(ReadSet1)});
+                            read_set=ReadSet1});
                         _ ->
                             ok = ?CLOCKSI_VNODE:commit(UpdatedPartitions, Transaction, CommitTime),
                             {next_state, receive_committed,
-                               S0#tx_coord_state{num_to_ack=NumToCommit, read_set=lists:reverse(ReadSet1), state=committing}}
+                               S0#tx_coord_state{num_to_ack=NumToCommit, read_set=ReadSet1, state=committing}}
                     end;
                 _ ->
                     {next_state, receive_prepared, S0#tx_coord_state{num_to_read= NumToRead-1,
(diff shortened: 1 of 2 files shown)


aletomsic:
Hi, In theory, the read_objects call does not use the static_tx_coordinator, so this changes in theory should not affect the result of your test.

1. Are you sure your test fails in master and does not fail in your branch (when run multiple times)?

2. I believe a better place to add a test like this one is in the clocksi_SUITE, given that it tests the behavior of transactions. I've just added a test in a branch I'm about to merge with master, and it passes (without the change). 
Could you please check that the following test is equivalent to what you're testing:

https://github.com/SyncFree/antidote/blob/parallel_updates/test/clocksi_SUITE.erl#L770-L779

If so, we can create an issue and mark it as resolved by that PR, too.
I can add a more explicit test name for that.

As I mention in Issue #248, I believe the static_coordinator should be completely removed. 
Check that out.

peterzeller:
The code I changed is used by the testcase, it failed before I did the change.

Annette just told me that she already fixed this a few months ago in the same file (https://github.com/SyncFree/antidote/commit/dc56a292bd0dd57cc0064c0f7d36ffaf1ec49e55) but she did not merge it.

My test works on your parallel_updates branch, so I am closing this and hope that #250 gets merged soon.

Your testcase tests the same thing as mine, but works on an interactive transaction. Not sure if that makes a difference...

aletomsic:
> On Nov 7, 2016, at 14:53, Peter Zeller <notifications@github.com> wrote:
> 
> The code I changed is used by the testcase, it failed before I did the change.
> 
> Annette just told me that she already fixed this a few months ago in the same file (dc56a29 <https://github.com/SyncFree/antidote/commit/dc56a292bd0dd57cc0064c0f7d36ffaf1ec49e55>) but she did not merge it.
> 
> My test works on your parallel_updates branch, so I am closing this and hope that #250 <https://github.com/SyncFree/antidote/pull/250> gets merged soon.
> 
> Your testcase tests the same thing as mine, but works on an interactive transaction. Not sure if that makes a difference…
> 
One of the things that my branch changes is that antidote.erl now calls the interactive txn coordinator instead of the static
when issuing an update_objects (which was before calling internally the deprecated interface clocksi_execute_tx). 

That’s why now it does not make a difference.

PR #255, opened 2016-11-07
Optional logging to disk

aletomsic:
This PR adds an option to disable writes to disk. 
This way, when benchmarking, the overhead of writes to disk log can be omitted.


src/logging_vnode.erl
```diff
@@ -235,11 +236,13 @@ init([Partition]) ->
 	    lager:error("ERROR: opening logs for partition ~w, reason ~w", [Partition, Reason]),
             {error, Reason};
         {Map,MaxVector} ->
+	    {ok, EnableLoggingToDisk} = application:get_env(antidote, enable_logging),
```
cmeiklejohn: How often will this be called?

test/log_recovery_SUITE.erl
```diff
@@ -62,33 +62,41 @@ all() -> [read_pncounter_log_recovery_test].
...
-    lager:info("Vnodes are started up"),
-    lager:info("Nodes: ~p", [Nodes]),
-
-    %% Read the value again
-    {ok, [ReadResult3], _CT} = rpc:call(FirstNode, antidote, read_objects,
-             [CommitTime, [], [Obj]]),
-    ?assertEqual(15, ReadResult3),
-    lager:info("read_pncounter_log_recovery_test finished").
+
+    case rpc:call(FirstNode, application, get_env, [antidote, enable_logging]) of
```
cmeiklejohn: Don't you have to handle the error case, incase it hasn't been set yet, since you aren't providing a default value?

deepthidevaki:
We have a config parameter sync_log, which toggles synchronous writes to disk. Do we need another parameter to turn of logging?

cmeiklejohn:
Does the other switch only switch between async and sync?

deepthidevaki:
Yes. 'sync_log = false' will write ops to disk, but asynchronously.

aletomsic:
The difference is that this completely removes writing to disk.
Besides, the disk_log:sync calls are independent of the sync or async writes.
Even when logging is set to async, the performance difference is important.

src/logging_vnode.erl
aletomsic: only when the vnode is initted.

test/log_recovery_SUITE.erl
aletomsic: I believe that case should not happen. 
As it is a config param that is set in antidote before it starts replying to rpcs.

src/inter_dc_sub_buf.erl
```diff
@@ -101,15 +105,23 @@ process_queue(State = #state{queue = Queue, last_observed_opid = Last}) ->
...
+			        lager:info("Whoops, lost message. New is ~p, last was ~p. Asking the remote DC ~p",
+				        [TxnLast, Last, State#state.pdcid]),
+			        case query(State#state.pdcid, State#state.last_observed_opid + 1, TxnLast) of
+				        ok ->
+					        State#state{state_name = buffering};
+				        _ ->
+					        lager:warning("Failed to send log query to DC, will retry on next ping message"),
+					        State#state{state_name = normal}
+			        end;
+		        false -> %% we deliver the transaction as we can't ask anything to the remote log
```
bieniusa: What is the "remote" log?

test/log_recovery_SUITE.erl
bieniusa: Shouldn't we just set a default value?

cmeiklejohn: You might want to handle `badrpc` explicitly here, otherwise if the RPC fails for some reason, you're going to crash.

src/inter_dc_sub_buf.erl
aletomsic: This checks that transactions arrive in total order from a DC to another.
Whenever a transaction arrives an "unknown" one, meanning there is a gap, one will ask the log of the remote DC to resend that transaction. 
When logging is disabled, this can't be done.

aletomsic: Maybe in the future we can create an issue to solve this in another way, but now I guess it would be good to have the option of disabling the logging for our benchmarks.

test/multiple_dcs_node_failure_SUITE.erl
```diff
@@ -83,7 +83,7 @@ cluster_failure_test(Config) ->
     case rpc:call(Node1, application, get_env, [antidote, enable_logging]) of
         {ok, false} ->
             pass;
-        {ok, true} ->
```
bieniusa: Why are you matching here against all possible returns?

aletomsic:
same answer here.

PR #256, opened 2016-11-10
More options for transactions

tcrain:
This extends the transactions with a few additional operations/features.

The first is that now transactions can take a list of properties, the properties currently supported are:

update_clock: If set to true the snapshot used for the transaction will be the most recent possible, if set to false it will use the vectorclock provided by the client.  Defaults to true.

certify: Whether or not to certify specifically this transaction.  Defaults to what is set in antidote.app.src

Additionally two operations are added to the interface of antidote

get_objects:  This operation is like the read except it returns the object state (i.e. the full CRDT object) instead of the value.

get_log_operations:  This operation takes a list of key/vectorclock pairs.  It then returns a list of all operations that are more recent then the given vectorclock for each key.  This is to be used along with the get_objects call so the client can keep the state of its objects up to date without fetching the entire state.

To do this the operation that reads from the log was made more flexible so that a range can be requested.  While doing this I updated the dialyzer specs for the dicts so they are now typed (this was added in the newer version of erlang).  This should make the typing more robust.

[Note that this was mostly code that was done a long time ago, but never merged, there will be a few more pull requests like this].


peterzeller:
> update_clock: If set to true the snapshot used for the transaction will be the most recent possible, if set to false it will use the vectorclock provided by the client. Defaults to true.

Does this mean that I can read arbitrary versions from the past, even ones that never were a recent snapshot on the current replica? 

And does it mean, that this snapshot is used as the context for Crdt operations? If so this would make it hard to add something like actor-ids for optimization later, I think. However, it would be nice to have,  since it could be used instead of interactive transactions in some cases.

tcrain:
Yes, it allows arbitrary versions to be read.

The snapshot is used as normal, so the reads will use that snapshot and the downstream generation will as well.

I agree that there are many possible downsides to allowing this, but it seems that there are many use cases for it as well.  Maybe I should add a flag in antidote.app.src that can disable it completely.

bieniusa:
I am about to fix the merge requests.

bieniusa:
@deepthidevaki Can you review the changes? 

There are two issues (that are not directly related to these changes):
1) The log format has substantially changed it seems. In the current format, I didn't see the operation name anymore - isn't this needed for replaying the operations?
2) Some tests seem to fail non-deterministically, but with the current error messages it is impossible to debug ( @mweberUKL I start to see your problems).

Should we add issues for these two items?

marc-shapiro:
Remember the log is Antidote's sole authoritative persistent memory.  If anything goes wrong we must be able to cold- or warm-restart from the log.  Implications:
Its format (etc.) must be PERFECTLY documented.  When something breaks in the future we must be able to fix it!
There should be regression tests, tools to read/write the log, verify the log invariants, etc.
As the system evolves we will have multiple log formats co-existing.  Each one must be self-identifying so that future versions of Antidote can read back old formats, and even mixed formats (progression from one format to the next might be non-monotonic).	

			Marc

deepthidevaki:
>
>
>
>    1. The log format has substantially changed it seems. In the current
>    format, I didn't see the operation name anymore - isn't this needed for
>    replaying the operations?
>
> log record consists of an update_payload which has an op. It is a
downstream update and antidote_crdt knows how to apply them.
https://github.com/SyncFree/antidote/blob/master/include/antidote.hrl#L115

bieniusa:
@marc-shapiro No worries, actually nothing regarding the log format has changed - I just got confused regarding the way the more recent optimised variant for the ORSet is representing operations. My problem was that I couldn't recall any changes to the log, just the message now looks different and I didn't make the link to the new set implementation.

The log format is documented by the type signature of the log entries.
If you want to start some documentation for the webpage, you are welcome :)

src/antidote.erl
```diff
@@ -52,6 +55,54 @@ start() ->
...
+        case clocksi_interactive_tx_coord_fsm:perform_singleitem_get(Clock, {Key, Bucket}, Type, Properties) of
+        {ok, Val, CommitTime} ->
+            get_objects_internal(Rest, Properties, [{Val, CommitTime}|Acc]);
+        {error, Reason} ->
+            {error, Reason}
+        end;
+    {error, Reason} ->
+        {error, Reason}
+    end.
+
```
bieniusa: Fixed that.

include/antidote.hrl
```diff
@@ -157,9 +157,12 @@
 
 -define(CLOCKSI_TIMEOUT, 1000).
 
+-type txn_properties() :: [{update_clock, boolean()} | {certify, use_default | certify | dont_certify}].
+
```
bieniusa: I would leave this for the moment being and refactor the properties at a later time. I don't understand fully yet how the txn_properties work in combination with the default parameter.

bieniusa:
@deepthidevaki Ready for pulling?

PR #258, opened 2016-11-15
(WIP) Improve error handling for keepalive connections

peterzeller:
@deepthidevaki did this together with me

Problem: When the client crashes during a transaction, the fsm is stuck in an already started transaction.
So when the client tries to start a new transaction, the fsm crashes and the connection is no longer  usable if it is KeepAlive=true.

Fix: When the clocksi_interactivetx_coord_fsm crashes, it will send an error to the process that started the transaction. If the connection is using KeepAlive it will try to restart the transaction with a new fsm.

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -193,6 +193,8 @@
 %% @doc Data Type: state
 %% where:
 %%    from: the pid of the calling process.
+%%    keepalive_from: the pid of the process, which started the transaction, 
+                      used for sending errors on crashes
 %%    txid: transaction id handled by this fsm, as defined in src/antidote.hrl.
 %%    updated_partitions: the partitions where update operations take place.
 %%    num_to_ack: when sending prepare_commit,
@@ -204,6 +206,7 @@
 
 -record(tx_coord_state, {
           from :: undefined | {pid(), term()},
+          keepalive_from :: pid() | undefined,
           transaction :: undefined | tx(),
           updated_partitions :: list(),
           client_ops :: list(), % list of upstream updates, used for post commit hooks
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -194,7 +194,7 @@
 %% where:
 %%    from: the pid of the calling process.
 %%    keepalive_from: the pid of the process, which started the transaction, 
-                      used for sending errors on crashes
+%%                    used for sending errors on crashes
 %%    txid: transaction id handled by this fsm, as defined in src/antidote.hrl.
 %%    updated_partitions: the partitions where update operations take place.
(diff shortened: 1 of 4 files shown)


marc-shapiro:
The fix appears fragile. What happens if the signal is lost? Duplicated?

                        Marc

peterzeller:
It is only the start-transaction message that is sent again, no operations are retried. The scenario is the following:

1. Client starts transaction 1
2. Client crashes during tx1, so tx1 is still open
3. Client tries to start tx2
4. The fsm handling tx1 receives a start-transaction request in a state where it does not expect one, so it crashes
5. (fixed in this pull-request) Client tries to start tx2 using a new fsm

An alternative solution would be to start a new fsm for every transaction directly. I am not sure if the optimization to reuse fsm is worth the trouble.

Currently the tests for this pull-request are not passing. I think we broke the transactions which are not keep-alive. Will look into it.

peterzeller:
Discussed this with @deepthidevaki and we came to the conclusion, that it's probably better to always start a new fsm for each transaction. Maybe we can also remove the supervisors for the fsms.

aletomsic:
> On Nov 15, 2016, at 13:51, Peter Zeller <notifications@github.com> wrote:
> 
> Discussed this with @deepthidevaki <https://github.com/deepthidevaki> and we came to the conclusion, that it's probably better to always start a new fsm for each transaction.
> 
I agree with this.
I’ve just made a simple bench that checks how expensive is to start a new coordinator when compared to reusing the existing one.
It turns out that finding the already existing coord by calling erlang:whereis() + sending a message to the coord to start a transaction takes more time aprox 30usec more on avg.) than simply starting a new one.
We could simply default to keepalive = false in case we want to keep that option for the future.
> Maybe we can also remove the supervisors for the fsms.
> 
> 
I don't know about that.

aletomsic:
> On Nov 15, 2016, at 15:11, Alejandro Z. Tomsic <aletomsic@gmail.com> wrote:
> 
> 
>> On Nov 15, 2016, at 13:51, Peter Zeller <notifications@github.com <mailto:notifications@github.com>> wrote:
>> 
>> Discussed this with @deepthidevaki <https://github.com/deepthidevaki> and we came to the conclusion, that it's probably better to always start a new fsm for each transaction.
>> 
> I agree with this.
> I’ve just made a simple bench that checks how expensive is to start a new coordinator when compared to reusing the existing one.
> It turns out that finding the already existing coord by calling erlang:whereis() + sending a message to the coord to start a transaction takes more time aprox 30usec more on avg.) than simply starting a new one.
> We could simply default to keepalive = false in case we want to keep that option for the future.

We should ask Tyler maybe, as he’s the one that created this keepalive option.

PR #261, opened 2016-11-18
Monitor staleness

deepthidevaki:
Changes:

- A process which periodically measures staleness from the vectorclock and store it to an exometer. This value can be read by monitoring tools
- a config parameter to turn on/off periodical measurement. It is turned off by default.

This changes may be useful in evaluation of antidote.


bieniusa:
The travis error is unrelated and has been seen in different branches too.

Issue #266, opened 2016-11-23
Read servers not started automatically on single instance


peterzeller:
If you `make console` and execute the following you get no reply:

```
(antidote@127.0.0.1)1> {ok, TxId} = antidote:start_transaction(ignore, []).
{ok,{tx_id,1479924617721329,<0.2441.0>}}
(antidote@127.0.0.1)2> antidote:read_objects([{<<"key">>, antidote_crdt_counter, <<"bucket">>}], TxId).
```

Workaround @aletomsic gave me: The read servers have to be started manually:

```
(antidote@127.0.0.1)1> inter_dc_manager:start_bg_processes(stable).
19:13:40.627 [info] Starting heartbeat sender timers
19:13:40.633 [info] Starting read servers
ok
(antidote@127.0.0.1)2> {ok, TxId} = antidote:start_transaction(ignore, []).
{ok,{tx_id,1479924927705981,<0.3741.0>}}
(antidote@127.0.0.1)3> antidote:read_objects([{<<"key">>, antidote_crdt_counter, <<"bucket">>}], TxId).
{ok,[0]}
```

The problem was introduced with https://github.com/SyncFree/antidote/pull/254 . The system tests don't catch it, because they start multiple machines, join them and start the required processes.

aletomsic:
I guess I can close this.

PR #267, opened 2016-11-24
automatically start read servers

aletomsic:
This PR addresses issue #266.
Added an option (which defaults to true) to start read_servers automatically.
When antidote is meant to be run in a cluster, this has to be set to false, and start the read servers afterwards.
This is the best idea I could come up with that does not penalize the interactive coordinator and, therefore, all transactions.
A more definitive and automatic way of solving this issue should be envisioned (maybe by more enlightened people ;).


peterzeller:
@bieniusa Can you have a look at this and merge? It would be important for testing things locally, for example using the Docker image.

src/antidote_app.erl
```diff
@@ -65,6 +65,13 @@ start(_StartType, _StartArgs) ->
                   ok
             end,
 
+            case application:get_env(antidote, auto_start_read_servers) of
+                {ok, true} ->
+                    %% start read servers
+                    inter_dc_manager:start_bg_processes(stable);
+                _->
+                    dont_start_read_servers
```
bieniusa: Why does this return an atom??

src/antidote.app.src
```diff
@@ -39,12 +39,15 @@
   %%                stored to disk (this is very slow in the current logging setup)
   %%        false -> all updates are sent to the operating system to be stored to disk (eventually), but are not guaranteed to be stored durably on disk
   %%                 when the reply is sent
+  %% %% auto_start_read_servers:
+  %%        true -> read servers will start automatically. It should be set to true when Antidote will be run in a single node/machine.
+  %%        false -> read servers will not start automatically. It should be set to false when Antidote will be run in a cluster, and later on start them.
```
bieniusa: This sentence I do not understand "and later on start them".

peterzeller:
> Can we have a systest that tests this code?

It is hard to test with our system tests. I created another pull request with an alternative way to test it ( #270 )

src/antidote_app.erl
aletomsic: this atom is not assigned to anything nor returned.
It is there just to "explain" that read servers won't be started.
In this branch of the case, nothing should be done.
Should I use an IF instead?

Issue #265, opened 2016-11-22
Nondeterministic system tests


peterzeller:
Sometimes the system tests are failing on Travis. 

Probably the test setup has some race condition.

Example logs of failing travis builds:

[antidote-fail-log1.txt](https://github.com/SyncFree/antidote/files/607008/antidote-fail-log1.txt)
[antidote-fail-log2.txt](https://github.com/SyncFree/antidote/files/607007/antidote-fail-log2.txt)
[antidote-fail-log3.txt](https://github.com/SyncFree/antidote/files/607006/antidote-fail-log3.txt)

The error always is something like this:

```
2016-11-22 16:21:50.874
16:21:50.873 [error] Error in process <0.1664.0> on node 'ct@testing-gce-9c7052de-389f-4c25-9d20-64660fe2fb35' with exit value:
{{ badmatch,{error,{antidote,
    {bad_return,{{antidote_app,start,[normal,[]]},
        {'EXIT',{{badmatch,{badrpc,{'EXIT',{noproc,{gen_server,call,
            [inter_dc_query,{del_dc,{'dev3@testing-gce-9c7052de-389f-4c25-9d20-64660fe2fb35',{1479,831551,546284}}},infinity]}}}}}
        ,[
        {inter_dc_manager,'-forget_dc/1-fun-2-',2,[{file,"/home/travis/build/SyncFree/antidote/_build/default/lib/antidote/src/inter_dc_manager.erl"},{line,233}]},
        {lists,foreach,2,[{file,"lists.erl"},{line,1337}]},
        {inter_dc_manager,forget_dc,1,[{file,"/home/travis/build/SyncFree/antidote/_build/default/lib/antidote/src/inter_dc_manager.erl"},{line,233}]},
        {lists,foreach,2,[{file,"lists.erl"},{line,1337}]},
        {inter_dc_manager,reconnect_dcs_after_restart,1,[{file,"/home/travis/build/SyncFree/antidote/_build/default/lib/antidote/src/inter_dc_manager.erl"},{line,197}]},
        {inter_dc_manager,check_node_restart,0,[{file,"/home/travis/build/SyncFree/antidote/_build/default/lib/antidote/src/inter_dc_manager.erl"},{line,187}]},
        {antidote_app,start,2,[{file,"/home/travis/build/SyncFree/antidote/_build/default/lib/antidote/src/antidote_app.erl"},{line,60}]},
[...]

PR #268, opened 2016-11-25
Fix crashing during test setup

tcrain:
This addresses the issue here:
https://github.com/SyncFree/antidote/issues/265

The problem that when a node restarted from an already existing configuration (i.e. restart after loading the state from disk), it would try to instruct all local nodes to reconnect to external DCs, instead of just itself.  Now when a node restarts it only tries to reconnect itself to the external DCs.


bieniusa:
Would it be possible to write a test for this?

PR #270, opened 2016-11-29
Release test script

peterzeller:
This pull request extends the travis script so that it builds a release and performs a minimal test on the release using a protocol-buffer transaction.

This test fails with a timeout because of bug  #266 and is fixed by pull request #267. It is hard to test this in the system tests, because the setup of the system tests already starts the required service.

I think testing the release in Travis is also useful in case someone changes the release scripts.

@bieniusa You might want to merge this into #267 first and then merge both into master together.

I also added the unit tests to the travis config.


test/release_test.escript
```diff
@@ -0,0 +1,44 @@
...
+    {ok, Pid} = try_connect(10),
+    % Try to read something:
+    Bound_object = {Key, antidote_crdt_counter, <<"release_test_key_bucket">>},
+    {ok, Tx2} = antidotec_pb:start_transaction(Pid, ignore, {}),
+    {ok, [Val]} = antidotec_pb:read_objects(Pid, [Bound_object], Tx2),
+    {ok, _} = antidotec_pb:commit_transaction(Pid, Tx2),
+    Value = antidotec_counter:value(Val),
+    true = Value >= 0,
+    _Disconnected = antidotec_pb_socket:stop(Pid),
+    io:format("Release is working, counter = ~p!~n", [Value]),
```
bieniusa: I would remove the counter value here, this is confusing. You already check that it has the right value anyway.

peterzeller:
I added some retries to the script and now it seems to work.

However, I think this shows some issues with Antidote:

- the PB port is started before it can answer requests within the timeout (10 seconds in antidotec_pb)
- the PB client connection seems to be broken after the first timeout. I have to reconnect after the timeout or it will timeout again. Maybe a deadlock somewhere.

PR #272, opened 2016-12-06
README.md: added simple pointer to our basho_bench branch

aletomsic:


bieniusa:
Now, if you add to the documentation how to run some basho-bench, view the results etc., this would be really helpful :)

aletomsic:
I added that in the readme inside basho bench. 
You reach it by following that link.

Alejandro Tomsic

PR #273, opened 2016-12-10
updated crdt libary and added pb-test for new map

bieniusa:


bieniusa:
@peterzeller Can you resolve the conflict and check/update the antidote_crdt dependency?

PR #275, opened 2016-12-30
Fix join local cluster scripts

ergl:


bieniusa:
@cmeiklejohn What about the mess support for Antidote? If this is not option anymore, we should clean up the scripts.

cmeiklejohn:
What is `mess`?

bieniusa:
@ergl What is the state of this pull request?
@cmeiklejohn What about the Mesos support for Antidote? If this is not option anymore, we should clean up the scripts.

ergl:
@bieniusa I haven't really touched this branch in a while, but apart from the changes to `bin/env`, I don't think this PR is needed anymore. I might be wrong on this, but it seems like using docker for local multi-node deployments is better than the current scripts in this branch.

PR #296, opened 2017-05-15
Operation compression

gmcabrita:
This pull request implements support for operation compression (taken from my non-uniform replication branch).

The basic idea is that when compression is enabled the `inter_dc_log_sender_vnode` will collect committed transactions for the duration specified in `antidote.hrl` instead of immediately broadcasting them. Once the timer ends the transactions are compressed and then broadcasted.

**This pull request depends on SyncFree/antidote_crdt#18**


albsch:
Closing and archiving this feature [here](https://antidotedb.gitbook.io/documentation/development/archived-works).

PR #302, opened 2017-06-20
Added description of Antidote-CRDTs

peterzeller:
I added a page describing the CRDTs in Antidote. Can someone have a look at this? Maybe Carlos, @gyounes and/or Annette.

- [ ] Add introduction with a brief explanation of how concurrent operations may occur and what is causally after.
- [ ] Add examples with figures for the different CRDTs
- [ ] Add code examples for Erlang, Javascript, Java clients (and maybe the low level rpc-calls and protocol buffers)

### DIFF
--- a/docs/pages/mydoc/crdts.md
+++ b/docs/pages/mydoc/crdts.md
@@ -12,7 +12,7 @@ Antidote offers a variety of other datatypes with smarter strategies to resolve 
 The available data types are described below.
 
 All data types have in common, that their value only depends on the operations which have been delivered to the currently used Antidote data center.
-For all datatypes there can be concurrent operations which are not yet visible.
+There can always be operations which are not yet visible, for example because they happened concurrently on a different datacenter.
 
 In the descriptions below the word "after" always means "causally after".
 An operation which happens later in time can be concurrent, if it happened independently.
@@ -55,7 +55,7 @@ reset()
 ### Counter
 
 The counter datatype only allows incrementing and decrementing the value.
-Reading the value returns the aggregated value from all operations.
+Reading the value returns the aggregated value from all received operations.
 
 #### Operations
 
@@ -64,9 +64,6 @@ increment(v)
 
 decrement(v)
 :	Decrements the counter by `v`.
-
-reset()
-:   Resets the register to the initial state (empty list).
 
 
 ### Fat Counter
@@ -85,7 +82,7 @@ decrement(v)
 :	Decrements the counter by `v`.
 
 reset()
-:   Resets the counter to the initial state (0).
+:   Resets the counter to the initial state (0), cancelling all observed increments and decrements.
 
 
 
@@ -173,7 +170,7 @@ reset()
 ### Grow-only Map (G-Map)
 
 The grow-only map does not support removing entries.
-It can be used for representing objects/structs in Antidote, where fields are dynamically added and removed.
+It can be used for representing objects/structs in Antidote, where fields are not dynamically added and removed.
 
 ### Add-wins Map (AW-Map)
 
@@ -186,7 +183,7 @@ The drawback of this implementation is, that markers for removed elements might 
 
 ### Remove-Resets Map (RR-Map)
 
-A `remove` on an remove-resets map resets the embedded datatype by calling its reset operation.
+Similar to the add-wins map, calling `remove` on a remove-resets map resets the embedded datatype by calling its reset operation.
 
 If the internal state of an embedded datatype becomes equal to its initial state, it is removed from the map.
 When calling `reset` this is the case for add-wins sets, multi-value registers, fat counters, flags and remove-reset maps which only contain the types in this sentence (read this as an inductive definition).
--- a/docs/pages/mydoc/crdts.md
+++ b/docs/pages/mydoc/crdts.md
@@ -1,0 +1,240 @@
+---
+title: Datatypes in Antidote
+tags: [crdts]
+sidebar: mydoc_sidebar
+permalink: crdts.html
+folder: mydoc
+---
+
+
+Most highly available databases only provide last-writer-wins registers, which resolve concurrent updates using timestamps.
+Antidote offers a variety of other datatypes with smarter strategies to resolve concurrent updates.
+The available data types are described below.
+
+All data types have in common, that their value only depends on the operations which have been delivered to the currently used Antidote data center.
+For all datatypes there can be concurrent operations which are not yet visible.
+
+In the descriptions below the word "after" always means "causally after".
+An operation which happens later in time can be concurrent, if it happened independently.
+
+Many datatypes support a `reset` operation, which is used when nesting a datatype in a map-datatype.
+Therefore the details of `reset` are explained with the map-datatypes below.
+
+
+
+## Last-writer-wins Register (LWW-Register)
+
+Reading a Last-writer-wins Register (LWW-Register) returns the latest written value based on timestamps.
+The timestamps respect causality, so assigning a new value is guaranteed to overwrite previous values.
+
+The initial value is the empty binary.
+
+#### Operations
+
+assign(value)
+:	Assigns a value to the register.
+
+## Multi-value Register (MV-Register)
+
+Reading a Multi-value Register (MV-Register) returns a list of all concurrently assigned values, which have not been overridden by other assignments.
+The returned list is ordered by the lexicographic order on the byte representation of the value.
+
+In the initial state reading the register returns the empty list.
+
+
+#### Operations
+
+assign(value)
+:	Assigns a value to the register.
+
+reset()
+:   Resets the register to the initial state (empty list).
+
+## Number Types
+
+### Counter
+
+The counter datatype only allows incrementing and decrementing the value.
+Reading the value returns the aggregated value from all operations.
+
+#### Operations
+
+increment(v)
+:	Increments the counter by `v`.
+
+decrement(v)
+:	Decrements the counter by `v`.
+
+reset()
+:   Resets the register to the initial state (empty list).
+
+
+### Fat Counter
+
+The fat counter offers the same functionality as the counter above but additionally supports the `reset` operation, which is useful for embedding the counter in a map datatype.
+
+The cost of this additional feature is an increased storage size, which is linear in the number of operations since the last reset.
+
+
+#### Operations
+
+increment(v)
+:	Increments the counter by `v`.
+
+decrement(v)
+:	Decrements the counter by `v`.
+
+reset()
+:   Resets the counter to the initial state (0).
+
+
+
+### Integer
+
+The integer extends the counter with an additional `set`-operation which allows to assign a value to the integer.
+
+Intuitively concurrent updates are resolved by taking the maximum in the case of concurrent `set`-operations and by considering all increments and decrement operations which are not overwritten by the `set`-operation.
+
+More precisely, the result of reading the value can be specified as follows:
+
+- If there are no `set` operations, the semantics are equivalent to the counter datatype.
+- Otherwise, let `S` be the set of all concurrent `set`-operations, which have not been overridden.
+- For each operation `o` in `S` with argument `assign(v)`, take all increments and decrements concurrent to `o` and aggregate them onto `V`.
+- Return the maximum of the values obtained from the previous step.
+
+#### Operations
+
+increment(v)
+:	Increments the integer by `v`.
+
+decrement(v)
+:	Decrements the integer by `v`.
+
+set(v)
+:   Sets the value of the integer to `v`.
+
+reset()
+:   Same as `set(0)`.
+	This is not a standard implementation of `reset` since  it might affect concurrent updates and does not reset the internal state to the initial state.
+
+## Flags
+
+Flags store a Boolean value (`true` or `false`).
+
+Antidote has two variants for flags, which differ in how concurrent updates are resolved.
+
+The **enable-wins flag** (EW-flag) returns true, if (and only if) there is an `enable`-operation among the latest operations.
+
+The **disable-wins** (DW-flag) flag return true, if (and only if) is an `enable`-operation and no `disable` operation among the latest operations.
+
+Here "latest operations" are all operations which have not been overridden by a later operation.
+
+
+#### Operations
+
+enable()
+:	Sets the flag to `true`.
+
+disable()
+:   Sets the flag to `false`
+
+reset()
+:   Resets the flag to the initial state (`false`).
+
+
+## Maps
+
+Maps in Antidote can store values under keys.
+The keys are simple (binary) values.
+Values are again datatypes and updating nested values is done by sending operations to these embedded values.
+Similarly, reading a value reads all the embedded values and returns them with their respective key.
+
+The different variations of maps differ in how they handle `remove`-operations.
+
+All implementations have in common, that they call `reset` on the removed value.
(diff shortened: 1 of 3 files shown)


preguica:
How can we check the final state of the page? Following the link I see
added lines, but no formatting....

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+
+
+Most highly available databases only provide last-writer-wins registers, which resolve concurrent updates using timestamps.
+Antidote offers a variety of other datatypes with smarter strategies to resolve concurrent updates.
+The available data types are described below.
+
+All data types have in common, that their value only depends on the operations which have been delivered to the currently used Antidote data center.
+For all datatypes there can be concurrent operations which are not yet visible.
+
+In the descriptions below the word "after" always means "causally after".
```
preguica: We needs images explaining what this means.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+An operation which happens later in time can be concurrent, if it happened independently.
+
+Many datatypes support a `reset` operation, which is used when nesting a datatype in a map-datatype.
+Therefore the details of `reset` are explained with the map-datatypes below.
```
preguica: It should explain here or later on if it is possible to nest a datatype that does not has a reset.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+
+
+
+## Last-writer-wins Register (LWW-Register)
+
+Reading a Last-writer-wins Register (LWW-Register) returns the latest written value based on timestamps.
+The timestamps respect causality, so assigning a new value is guaranteed to overwrite previous values.
```
preguica: overwrite **observed** values.

what does previous means?

peterzeller:
> How can we check the final state of the page? Following the link I see added lines, but no formatting....

Hmm, I did not find a way to get a preview of the page in a pull-request, except for installing jeykll locally. 

There is a preview of the Markdown here, but it does not show the final rendering: https://github.com/SyncFree/antidote/blob/crdt-documentation/docs/pages/mydoc/crdts.md

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### Operations
+
+assign(value)
+:	Assigns a value to the register.
+
+## Multi-value Register (MV-Register)
+
+Reading a Multi-value Register (MV-Register) returns a list of all concurrently assigned values, which have not been overridden by other assignments.
+The returned list is ordered by the lexicographic order on the byte representation of the value.
```
preguica: why? is there any good reason for this?

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+
+reset()
+:   Resets the register to the initial state (empty list).
+
+## Number Types
+
+### Counter
+
+The counter datatype only allows incrementing and decrementing the value.
+Reading the value returns the aggregated value from all operations.
```
preguica: all received operations.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+decrement(v)
+:	Decrements the counter by `v`.
+
+reset()
+:   Resets the register to the initial state (empty list).
+
+
+### Fat Counter
+
+The fat counter offers the same functionality as the counter above but additionally supports the `reset` operation, which is useful for embedding the counter in a map datatype.
```
preguica: Does not explain the interaction between reset and inc/dec.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+reset()
+:   Resets the counter to the initial state (0).
+
+
+
+### Integer
+
+The integer extends the counter with an additional `set`-operation which allows to assign a value to the integer.
+
+Intuitively concurrent updates are resolved by taking the maximum in the case of concurrent `set`-operations and by considering all increments and decrement operations which are not overwritten by the `set`-operation.
```
preguica: Just to understand...  if we start with a counter with value 1 and concurrently we do:

rep1: inc, inc , inc, set(2)

rep2: set(2)

we end up with 5, right? This does not make sense to me.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+:   Same as `set(0)`.
+	This is not a standard implementation of `reset` since  it might affect concurrent updates and does not reset the internal state to the initial state.
+
+## Flags
+
+Flags store a Boolean value (`true` or `false`).
+
+Antidote has two variants for flags, which differ in how concurrent updates are resolved.
+
+The **enable-wins flag** (EW-flag) returns true, if (and only if) there is an `enable`-operation among the latest operations.
```
preguica: have you defined latest operations?

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+A `remove` on an add-wins map resets the embedded datatype by calling its reset operation.
+Moreover, it marks the entry as removed, so that it does not show up when reading the value of the map.
+
+Concurrent updates on a key will overwrite the removed-marker (i.e. adds via update-operation win over concurrent removes, hence the name).
+
+The drawback of this implementation is, that markers for removed elements might never be cleaned up, so it should only be used when the set of keys is limited.
+
+### Remove-Resets Map (RR-Map)
+
+A `remove` on an remove-resets map resets the embedded datatype by calling its reset operation.
```
preguica: this is the exact same sentence with add-wins map.

we need examples on this one.

gyounes:
Hi,

I can do that :)

Best,
Georges

preguica:
I have left some comments there. I have some time to help writing this in
the following days/weeks.

In my opinion, this should start with a brief explanation of how concurrent
operations may occur and what is causally after.
Then, we need examples with each data type.

Maybe we could even try to do the same as Riak had, with code snippets for
the use of operations in different languages.

We can sync on this tomorrow or thursday.

preguica:
On Tue, Jun 20, 2017 at 11:45 PM, Nuno Preguica <nuno.preguica@fct.unl.pt>
wrote:

> I have left some comments there. I have some time to help writing this in
> the following days/weeks.
>
> In my opinion, this should start with a brief explanation of how
> concurrent operations may occur and what is causally after.
>

With a explanation, I mean figures showing what can happen that leads to
concurrent operations.

gyounes:
It is a great idea, explaining using the figures.
I can start with adding/improving the explanation for now and work on the
figures after discussing this (whether tomorrow or thursday)

docs/pages/mydoc/crdts.md
preguica: set should win over concurrent inc/dec

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+folder: mydoc
+---
+
+
+Most highly available databases only provide last-writer-wins registers, which resolve concurrent updates using timestamps.
+Antidote offers a variety of other datatypes with smarter strategies to resolve concurrent updates.
+The available data types are described below.
+
+All data types have in common, that their value only depends on the operations which have been delivered to the currently used Antidote data center.
+For all datatypes there can be concurrent operations which are not yet visible.
```
CBaquero: Only concurrent operations may not yet be visible? There could also be operations that are in the future of the current state and not yet visible. If so then this should write as "For all datatypes there can be concurrent or following operations which are not yet visible."

docs/pages/mydoc/crdts.md
CBaquero: I think it makes sense to nest datatypes that do not have a reset only for grow-only maps and other grow-only compositions. Reset is used only when supporting deletion of elements.

docs/pages/mydoc/crdts.md
CBaquero: Agree

CBaquero: If there is only one value at each time, can we say it overwrites previous values in plural?

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+#### Operations
+
+increment(v)
+:	Increments the counter by `v`.
+
+decrement(v)
+:	Decrements the counter by `v`.
+
+reset()
+:   Resets the register to the initial state (empty list).
```
CBaquero: This counter has a reset, but will probably not compose well inside non grow-only maps. Maybe note that.

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+#### Operations
+
+increment(v)
+:	Increments the counter by `v`.
+
+decrement(v)
+:	Decrements the counter by `v`.
+
+reset()
+:   Resets the counter to the initial state (0).
```
CBaquero: "Resets the counter to the initial state (0), cancelling all observed increments and decrements."

docs/pages/mydoc/crdts.md
```diff
@@ -0,0 +1,240 @@
...
+:   Removes entries from the map based on keys.
+
+reset()
+:   Resets the map.
+
+
+### Grow-only Map (G-Map)
+
+The grow-only map does not support removing entries.
+It can be used for representing objects/structs in Antidote, where fields are dynamically added and removed.
```
CBaquero: If it can only grow, how are "fields are dynamically added and removed"

docs/pages/mydoc/crdts.md
peterzeller: I think "observed" sounds like you have to read the value before the write. 

For the LWW register a new write wins over previous no matter if you interpret "previous" as "smaller timestamp" or "happened before wrt. causality". Maybe we can make this clearer if we explain the concepts of causality in more detail in the introduction.

Regarding "previous values in plural": I was thinking about all the previous operations, but you are right that it is strange to say this for values.

docs/pages/mydoc/crdts.md
peterzeller: Erlang (and protocol buffers) do not have native sets, only lists. So to get convergence we need to fix some order for the result.

docs/pages/mydoc/crdts.md
peterzeller: Your example is correct. I guess the problem with letting `set` win is the same as with `reset`: to implement it we need to do something like with fat-counters.

So unless we add something more on the infrastructure level (e.g. causal stability as @gyounes presented in his Papoc paper) I would guess it is not possible to implement this efficiently. 

If we want to discuss this further we should probably move the discussion to some other issue or mailing list and discuss here only the description of the existing CRDTs. We can also remove the integer from the official interface, if it is too controversial.

docs/pages/mydoc/crdts.md
peterzeller: I would keep the repetition since I expect people to read the documentation only for single CRDTs and not from top to bottom. Maybe it can be formulated a bit nicer, though.

docs/pages/mydoc/crdts.md
peterzeller: When writing I did not think of the state but of the currently executing operation or the current transaction and since this is not complete yet, there can be no operations after it.

When thinking about the state I see that my sentence is not correct. I'll try to reformulate it.

docs/pages/mydoc/crdts.md
peterzeller: That was a copy-paste error, the normal counter does not report `reset`.

docs/pages/mydoc/crdts.md
peterzeller: I missed a "not" in that sentence ;)

The idea is to use it for something like a `user` struct with `name`, `mail`, `passwordhash` and other fields, where the set of fields usually does not change.

preguica:
On Thu, Jun 22, 2017 at 1:27 PM, Peter Zeller <notifications@github.com>
wrote:

> *@peterzeller* commented on this pull request.
> ------------------------------
>
> In docs/pages/mydoc/crdts.md
> <https://github.com/SyncFree/antidote/pull/302#discussion_r123495849>:
>
> > +
> +All data types have in common, that their value only depends on the operations which have been delivered to the currently used Antidote data center.
> +For all datatypes there can be concurrent operations which are not yet visible.
> +
> +In the descriptions below the word "after" always means "causally after".
> +An operation which happens later in time can be concurrent, if it happened independently.
> +
> +Many datatypes support a `reset` operation, which is used when nesting a datatype in a map-datatype.
> +Therefore the details of `reset` are explained with the map-datatypes below.
> +
> +
> +
> +## Last-writer-wins Register (LWW-Register)
> +
> +Reading a Last-writer-wins Register (LWW-Register) returns the latest written value based on timestamps.
> +The timestamps respect causality, so assigning a new value is guaranteed to overwrite previous values.
>
> I think "observed" sounds like you have to read the value before the write.
>
> For the LWW register a new write wins over previous no matter if you
> interpret "previous" as "smaller timestamp" or "happened before wrt.
> causality". Maybe we can make this clearer if we explain the concepts of
> causality in more detail in the introduction.
>
Makes sense.
Of course, smaller timestamp does not necessarily mean previous if clocks
are skewed... ;-)

preguica:
On Thu, Jun 22, 2017 at 1:35 PM, Peter Zeller <notifications@github.com>
wrote:

> *@peterzeller* commented on this pull request.
> ------------------------------
>
> In docs/pages/mydoc/crdts.md
> <https://github.com/SyncFree/antidote/pull/302#discussion_r123497377>:
>
> > +increment(v)
> +:	Increments the counter by `v`.
> +
> +decrement(v)
> +:	Decrements the counter by `v`.
> +
> +reset()
> +:   Resets the counter to the initial state (0).
> +
> +
> +
> +### Integer
> +
> +The integer extends the counter with an additional `set`-operation which allows to assign a value to the integer.
> +
> +Intuitively concurrent updates are resolved by taking the maximum in the case of concurrent `set`-operations and by considering all increments and decrement operations which are not overwritten by the `set`-operation.
>
> Your example is correct. I guess the problem with letting set win is the
> same as with reset: to implement it we need to do something like with
> fat-counters.
>
> So unless we add something more on the infrastructure level (e.g. causal
> stability as @gyounes <https://github.com/gyounes> presented in his Papoc
> paper) I would guess it is not possible to implement this efficiently.
>
I think what we need to add is ids for nodes... with that, it should be
possible to implement reset efficiently.
The problem of causal stability is that when you have client that can keep
object for a long time (or even cache objects), it becomes a mess.

bieniusa:
I suggest adding pointers to papers and the Syncfree webpage resources on CRDTs.

Also it would be great to have a small tutorial on CRDTs in Antidote.
Chris and I made a CRDT tutorial based on the Lasp CRDT library, but we should also be able to do that with the Antidote CRDTs.

pviotti:
What's the state of this pull request? It might not be perfect but I reckon it would be good if we could merge even this first version to master.

Some comments/questions:
- why are read operations not mentioned at all?
- it would be useful to link each datatype to its corresponding implementation in `rawapi.html`
- (there's an incomplete sentence: "Sends a list of nested operations to the")
- it would be useful to have a brief summary of what "embedding" means, and what can be embedded into what
- some parts of the prose sound a bit too "academic" and hard to parse for end-user documentation (e.g. the "inductive definition" in the intro of the RR-Map section)

In addition to this page, eventually, I think it would be helpful to have a sort of live playground in which the user can try animated CRDTs demos (similar to [this]( http://thesecretlivesofdata.com/raft/) but with user input).

PR #303, opened 2017-07-07
Fix vector_orddict:filter/2 returning wrong size, add tests

ergl:


bieniusa:
Hmmm, unfortunately there is no documentation of the expected behaviour of filter.
Returning both the result and the length of the result seems a bit superfluous to me: If you have the result, you can easily compute its length.

@peterzeller What is your take?

peterzeller:
I only appear in the `git blame`, because I moved the file from the `antidote_utils` project. The original source is here: https://github.com/SyncFree/antidote_utils/blame/master/src/vector_orddict.erl

> Returning both the result and the length of the result seems a bit superfluous to me: If you have the result, you can easily compute its length.

The size is used to determine when to do garbage collection. Computing the size every time would require iterating over the list, so it might be worth caching the size. My guess would be that it provides no measurable performance benefit.

Akso, I think this module would become obsolete if we integrate the backend from @santialvarezcolombo someday ;) 
So I would just accept this fix now and hope that we get a cleaner solution eventually.

Btw: I could not find any use of the filter-function in Antidote. Maybe we can just remove it?

PR #311, opened 2017-08-18
add  erlang 20 for test

getong:


bieniusa:
The code will break under Erlang 20. It needs quite a bit of refactoring that we will start once we figured who will be affected by the changes.

albsch:
The Erlang 20 branch is currently being merged.

PR #324, opened 2018-02-03
fix erlzmq dependency

pviotti:
There appears to be a problem with the zmq dependency: it fails when dowloading a file from http://pkgs.fedoraproject.org (see the [Travis job.](https://travis-ci.org/SyncFree/antidote/jobs/335612622)).  
This commit restores the dependency to its original repository, which uses another URL.


bieniusa:
Thanks!

PR #331, opened 2018-03-26
fix default epmd port in dockerfile

deepthidevaki:


bieniusa:
For documentation: Why is this change needed?

deepthidevaki:
The default port used by epmd is 4369 which was not exposed in the current docker image. Because of this, external erlang nodes cannot connect to antidote node running in the container.

PR #333, opened 2018-03-28
Towards a new materializer

bieniusa:
This is mainly cleaning a few things on the way.
- Removing unused modules
- Refactoring some tests
- Associating functionality with the correct module


src/materializer_vnode.erl
```diff
@@ -335,8 +272,65 @@ terminate(_Reason, _State=#mat_state{ops_cache=OpsCache, snapshot_cache=Snapshot
...
+            {error, Reason};
+        {NewContinuation, NewOps, OpsDict} ->
+            load_ops(OpsDict, State),
+            loop_until_loaded(Node, LogId, NewContinuation, NewOps, State);
+        {eof, OpsDict} ->
+            load_ops(OpsDict, State),
+            ok
+    end.
+
+-spec load_ops(dict:dict(key(), [{non_neg_integer(), clocksi_payload()}]), #state{}) -> true.
```
peterzeller: The honest return type would be `true | ok`, because `lists:foreach` returns `ok`

src/materializer_vnode.erl
```diff
@@ -634,86 +594,51 @@ check_filter(Fun, Id, Last, NewId, Tuple, NewSize, NewOps) ->
...
+%% 4) size of the tuple that contains the op list, 5) the tuple containing the op list
+%% Note that the ops in the tuple are stored in the ets with the most recent op at the end of
 %% the tuple.
-%% The second argument if true will convert the ops tuple to a list of ops
-%% Otherwise it will be kept as a tuple
--spec tuple_to_key(tuple(), boolean()) -> {any(), integer(), non_neg_integer(), non_neg_integer(),
-                      [op_and_id()]|tuple()}.
-tuple_to_key(Tuple, ToList) ->
+-spec deconstruct_opscache_entry(tuple()) ->
+    {key(), non_neg_integer(), non_neg_integer(), non_neg_integer(), [op_and_id()] | tuple()}.
```
peterzeller: The last component of the return type should be just `tuple()` instead of `[op_and_id()] | tuple()`. `[op_and_id()]` was only for the `true` case in the old code.

Probably this change also affects the types of `fetch_updates_from_cache` and the `ops_list` field in the `snapshot_get_response` record.

Maybe it would be clearer to write getters for the 4 parts?

src/antidote.app.src
```diff
@@ -15,6 +15,8 @@
                   antidote_pb,
                   erlzmq,
                   rand_compat,
+                  runtime_tools,
+                  tools,
```
peterzeller: Should this be here or was this just for debugging?

bieniusa: This was for debugging, but I don't see why they shouldn't be there in general.
The documentation says:
```[...]provides low footprint tracing/debugging tools suitable for inclusion in a production system.```

src/materializer_vnode.erl
bieniusa: I am hesitant to have more changes on materializer_vnode as this will be replaced anyway.

Issue #329, opened 2018-03-21
Build on Erlang 18 broken


bieniusa:
The build on Erlang 18 is broken due to the dependency on a version of Poolboy that is not compatible with Erlang 18.

Do we need to work on Erlang 18??

PR #340, opened 2018-04-24
Updated travis setup

bieniusa:
Removed Erlang 18 as moving from gen_fsm to gen_statem will break the
builds
Updated contact emails for notifications

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: erlang
 otp_release:
-  - 18.3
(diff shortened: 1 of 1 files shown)


bieniusa:
This fixes issue #329

goncalotomas:
Travis in the long term may only keep the latest release of each major version, in this case 19.3. Perhaps this would be a good time to replace 19.1 with 19.3?

PR #343, opened 2018-05-09
Don't ignore rebar.lock

ergl:


albsch:
Fixed with Pull Request #344.

PR #358, opened 2018-11-20
Dockerfile for a tiny container image of AntidoteDB based on Alpine Linux

ilyasToumlilt:
A Dockerfile for a tiny container image (3 times smaller than the default one) of AntidoteDB based on Alpine Linux.

+ This image is used for the Antidote playground
+ Also fixed the Github link in the default Dockerfile

### DIFF
--- a/Dockerfiles/Dockerfile-alpine
+++ b/Dockerfiles/Dockerfile-alpine
@@ -1,4 +1,5 @@
-FROM erlang:19
+# A Docker file for a tiny container image of AntidoteDB based on Alpine Linux
+FROM erlang:20.3.8-alpine
 
 ENV HANDOFF_PORT "8099"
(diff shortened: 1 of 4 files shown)


bieniusa:
Thanks, this looks neat.
Question: The version on Dockerhub will continue using the old image, right?
Should we change that?

@mweberUKL Any hint/comment from your side?

ilyasToumlilt:
Hi,

Yes, the Dockerhub version will continue to be built using the old Dockerfile, were I just updated the link to the git repo.

We should keep both versions on Dockerhub, with different tags. (like they do in https://hub.docker.com/_/erlang/ <https://hub.docker.com/_/erlang/>)
One version from the default setup, and a lightweight one (the Dockefile I added) with -alpine tag.

Ilyas

PR #365, opened 2019-03-25
Replaced internal riak_core functions with equivalent public ones

albsch:
Replace calls to `riak_core` module `app_helper` with equivalent `application` module calls.


albsch:
Ah, pushed to wrong branch...

PR #367, opened 2019-04-24
Allows bcounters to use standard counter inc op

SmallEndian:
This way bcounters can be used as with the same operations as antidotec_counter


bieniusa:
Can you please add some tests to make sure that the new variants are covered by the test suite?

bieniusa:
Ping! What is the status of this PR?

FairPlayer4:
@bieniusa 
Should I include this functionality in the bcounter_mgr?
I think this can also be added for the decrement operation and would allow the user to perform bcounter operations locally without knowing (or rather specifying) the local DC

albsch:
`bcounter` isn't a drop-in substitute for `antidote_crdt_counter_pn` at the API level, and unifying the op shape tends to hide that.

PR #369, opened 2019-05-10
Update LICENSE to standard format

ilyasToumlilt:
A random (but useful) PR to test the new Docker Hub auto-build.


bieniusa:
This pull request is not useful - the appendix is *not* part of the license nor is "antidotedb.eu" the copyright holder.
We have checked our licensing with legal departments to be compliant. I don't see any reason to change it.

PR #372, opened 2019-06-18
Cluster management pb fix

peterzeller:
Some unrelated changes, probably I should clean this up before merging.

- The `fix createDc function ` commit fixes a bug we introduced in our last pull-request because we changed the registered name for that process.
- Updating the pb library should fix another bug with cluster management. 
- The `pb_client_cluster_management_SUITE` is a new test for testing the cluster management. Apparently that is not tested at the moment. I still need to test joining 2 clusters. We should probably also add these functions to the Erlang client so that the test code is a bit nicer.

The other changes change the testing framework, so that it is easier to debug.


rebar.config
```diff
@@ -9,10 +11,10 @@
 
     %% antidote utilities
     antidote_crdt,
-    antidote_pb_codec,
```
bieniusa: Before committing, please check with the most recent pb_codec version!

test/singledc/pb_client_SUITE.erl
```diff
@@ -116,7 +116,7 @@ get_empty_crdt_test(_Config) ->
     Bound_object = {<<"key1">>, antidote_crdt_counter_pn, Bucket},
 
     {ok, Pid} = antidotec_pb_socket:start(?ADDRESS, ?PORT),
-    {ok, TxId} = antidotec_pb:start_transaction(Pid, ignore, []),
+    {ok, TxId} = antidotec_pb:start_transaction(Pid, term_to_binary(ignore), []),
```
peterzeller: Would probably be cleaner to call `term_to_binary` in the antidote client code.

src/antidote_pb_process.erl
```diff
@@ -37,24 +37,27 @@
 
 -export([process/1]).
 
--spec decode_clock(binary()) -> snapshot_time() | ignore.
+-spec decode_clock(binary()) -> snapshot_time() | ignore | txid().
```
peterzeller: The function name `decode_clock` is a bit misleading if it also handles  `txid`?

test/singledc/pb_client_SUITE.erl
bieniusa: done

src/antidote_pb_process.erl
```diff
@@ -64,39 +67,39 @@ process({abort_transaction, TxId}) ->
...
     end;
 
 
-process({static_read_objects, {Clock, Properties, Objects}}) ->
+process({static_read_objects, Clock, Properties, Objects}) ->
     Response = antidote:read_objects(decode_clock(Clock), Properties, Objects),
     case Response of
-        {error, Reason} ->{commit_response, {error, Reason}};
-        {ok, Results, CommitTime} -> {static_read_objects_response, {ok, lists:zip(Objects, Results), CommitTime}}
+        {error, Reason} -> {error_response, {error, Reason}};
```
peterzeller: Is changing this response backwards compatible for the Antidote clients?

src/antidote_pb_process.erl
```diff
@@ -106,30 +109,33 @@ process({create_dc, NodeNames}) ->
     catch
      Error:Reason -> %% Some error, return unsuccess. TODO: correct error response
        logger:info("Create DC Failed ~p : ~p", [Error, Reason]),
-       {operation_response, {error, create_dc_failed}}
+       {operation_response, {error, unknown}}
     end;
 
-process({get_connection_descriptor}) ->
+process(get_connection_descriptor) ->
```
peterzeller: :+1: for removing 1-tuples

src/antidote_pb_process.erl
bieniusa: Renamed to from_bin

PR #378, opened 2019-07-29
Antidote channels

balegas:
Replacement for inter-dc zeromq implementation.

Provides a library to abstract zeromq communication patterns from inter-dc replication code.
Supports ZeroMQ and RabbitMQ and provides an extensible interface.

Current implementation was tested for ZeroMQ. Performance is comparable to current implementation. In the long-term we will want to use this library to test the performance of RabbitMQ against ZeroMQ, or to replace the underlying erlzmq library.


FairPlayer4:
This will have to be updated significantly due to the changes in master
I like the idea of having an alternative message queue service

PR #376, opened 2019-07-25
Logging refactor

bieniusa:
- Removed outdated sasl config
- Adding logging config for disk and console

PR #379, opened 2019-07-30
rebar version 3.11.1

albsch:
Rebar executable version bump to get the logging to work properly


peterzeller:
```
# src/logging_vnode.erl [FAIL]
  - god_modules
    - This module has too many functions (35). Consider breaking it into a number of modules.
```

You should add the config change from the other branch before merging.

albsch:
Yes, let's wait until #376 is merged.

PR #382, opened 2019-08-12
Modify make command to enable local docker builds

albsch:
The docker build commands now reside in the `docker-antidote` repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,10 +92,3 @@ xref: compile
 
 dialyzer:
 	${REBAR} dialyzer
-
-docker-build:
-	docker build -f Dockerfiles/Dockerfile -t antidotedb/antidote Dockerfiles
-
-docker-local:
-	docker run --rm -v $(shell pwd):/code -w /code erlang:21 make rel
-	docker build -f Dockerfiles/Dockerfile-local -t antidotedb/antidote:local .


bieniusa:
I think we make the workflow unnecessarily complex for people who use local builds if we require to pull another repo. I was under the impression that the Docker repo was mainly intended for the Dockerhub Builds. Can you confirm @ilyasToumlilt  ?

albsch:
Also, why is `rebar3 as test coveralls send` failing? I don't see an error message.

bieniusa:
Coveralls is currently down for maintenance according to their webpage.

Also have a look at [that:](https://github.com/marketplace/actions/coveralls-github-action)

ilyasToumlilt:
> I think we make the workflow unnecessarily complex for people who use local builds if we require to pull another repo. I was under the impression that the Docker repo was mainly intended for the Dockerhub Builds. Can you confirm @ilyasToumlilt ?

Docker repo is mainly used for Dockerhub builds, but it contains good documentation for local ones also.
It makes sense for me to keep all Dockerfiles organised, updated and documented in the Docker repo, as we have many build recipes for different Antidote versions/dependencies.

I think, keeping this local build command can benefit to Antidote developers that want to quickly build a docker image from their modified sources without fetching the Docker repo and doesn't need Docker repo's benefits. But for Antidote users, it's better not to use this, and pull stable images from the hub.

albsch:
I'm going to modify the current make command in the antidote repository to enable local builds without pulling the `docker-antidote` repository.

PR #387, opened 2019-09-20
Implementation of strong consistency using locks

peterzeller:
This PR adds support for strongly consistent (serializable) transactions by using Locks.

Current status: 

There is still one failing test case:
The `cluster_failure_test_2` sometimes fails. The test crashes a node holding the locks and then restarts it after some time. After the restart, another DC should be able to get the locks again, but the crashed DC sometimes is not able to read its CRDT state after restarting. I still have to debug this case - I am not yet sure if this is a problem with the lock implementation or a general problem in Antidote. Unfortunately, the bug occurs less frequently the more log messages I add for debugging.

The rest of the implementation should be working. 

Documentation is available in https://github.com/AntidoteDB/antidote/blob/strong-consistency-3/src/antidote_locks.md


bieniusa:
From the documentation:

> The call to start_transaction may fail if Antidote is unable to acquire the lock.

What message is then sent to the client?

include/antidote_locks.hrl
```diff
@@ -0,0 +1,58 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright <2013-2018> <
```
bieniusa: Shouldn't this become 2019?

src/antidote_lock_server.erl
```diff
@@ -0,0 +1,634 @@
...
+%%    spawn_link(fun() ->
+%%        check_lock_state_process(Self)
+%%    end),
+    S = antidote_lock_server_state:initial(MyDcId, AllDcs, ?MIN_EXCLUSIVE_LOCK_DURATION, ?MAX_LOCK_HOLD_DURATION, ?INTER_DC_LOCK_REQUEST_DELAY),
+    {ok, #state{
+        s = S,
+        read_write_process = spawn_link(fun() -> read_write_process(Self) end)
+    }}.
+
+%%check_lock_state_process(Pid) ->
```
bieniusa: Can be deleted?

src/antidote_lock_server.erl
```diff
@@ -0,0 +1,634 @@
...
+    {ok, Values, ReadClock} = antidote:read_objects(Clock, [], Objects),
+    antidote_lock_server_state:debug_log({event, read_crdt_done, #{
+        clock => Clock,
+        objects => Objects,
+        values => Values,
+        cont => Cont
+    }}),
+    gen_server:call(Self, #on_read_crdt_state{cont = Cont, clock = ReadClock, values = Values}),
+    read_write_process(Self).
+
```
bieniusa: This file probably needs some final clean up?

src/antidote_locks.md
```diff
@@ -0,0 +1,120 @@
...
+    
+
+
+
+Future Work
+-----------
+
+- *Dynamic Membership*: Adding and removing DCs is currently not supported.
+- *Fault tolerance*: In the following scenarios, the current implementation will not work correctly:
+    - CRDT state cannot be updated locally: Implementation will wait for the update indefinitely.
```
bieniusa: I don't understand this comment: Which CRDT state update is blocking here?

src/clocksi_interactive_coord.erl
```diff
@@ -234,7 +234,8 @@ finish_op(From, Key, Result) ->
     return_accumulator :: list() | ok | {error, reason()},
     is_static :: boolean(),
     full_commit :: boolean(),
-    properties :: txn_properties()
+    properties :: txn_properties(),
+    locks :: undefined | antidote_locks:lock_spec()
```
bieniusa: Does `undefined` here mean `none`?

src/antidote_lock_server_state.erl
```diff
@@ -0,0 +1,1428 @@
...
+    ]),
+    Actions = [
+        #send_inter_dc_message{
+            receiver = Dc, message = #lock_request{locks = LockChanges, request_time = T, requester_pid = Pid}}
+        || Dc <- other_dcs(State), {Pid, T, LockChanges} <- Changes
+    ],
+    {Actions, NewState}.
+
+
+%%% computes lock request actions for the locally waiting locks
```
bieniusa: Is this still required?

src/antidote_locks.md
peterzeller: When a lock is updated I am using an CRDT update operation. This update might fail, e.g. if the partition with the lock is currently not available. There is no error handling for this case yet.

include/antidote_locks.hrl
peterzeller: It's the same in other files. I can create a separate PR to update the date everywhere.

albsch:
Using `-erl_args -hidden` causes every suite to restart and reconnect every node. This causes an infinite request `log_read` loop in the `clocksi_SUITE`, which indicates either a problem with the `hidden` parameter or an inter-dc communication problem (which goes back even before my inter-dc simpliification).

PR #388, opened 2019-10-01
Build a local Docker image from the source code

ilyasToumlilt:
+ For dev reasons we still need a way to build a local image from the source code
+ (portability) when benchmarking it's easier to build this image without worrying about dependencies


albsch:
I think we can move the Dockerfile to the docker-antidote repository where we decided all Dockerfiles should reside.

Then we only need this part merged into the Antidote repo, if we suppose that the local Dockerfile can be found in the `docker-antidote`:
```
docker-build:
    tmpdir=`mktemp -d` ; \
    wget "https://raw.githubusercontent.com/AntidoteDB/docker-antidote/master/local/Dockerfile" -O "$$tmpdir/Dockerfile" ; \
    wget "https://raw.githubusercontent.com/AntidoteDB/docker-antidote/master/local/entrypoint.sh" -O "$$tmpdir/entrypoint.sh" ; \
    wget "https://raw.githubusercontent.com/AntidoteDB/docker-antidote/master/local/start_and_attach.sh" -O "$$tmpdir/start_and_attach.sh" ; \
    cd $$tmpdir ; \
    docker build -f Dockerfile -t antidotedb:local-build .

docker-run: docker-build:
    docker run -d --name antidote -p "8087:8087" antidotedb:local-build
```

bieniusa:
Status?

ilyasToumlilt:
Update:
- Moved Dockerfile & scripts to https://github.com/ilyasToumlilt/docker-antidote/tree/master/local-build 
- Set the Makefile to fetch them as suggested by @albsch , with a hack to put the mktemp build dir inside de local directory so it can be inside the scope of docker build (Docker can only import files that are inside the directory where it runs from).
- Also added a docker-clean rule to the makefile.

PR #397, opened 2019-10-31
Antidote configuration

albsch:
Unified and simplified Antidote configuration.

* Changed shell command to use the release shell instead
* Added part on how to use the relase in the Readme
* Rebar now understands symlinks for _checkouts, fixed readme
* Removed deprecated env entrypoint script
* Removed launch-nodes script, uses legacy RELX_REPLACE_OS_VARS functionality
* Removed all schema files except the riak_core schema for compatibility purposes. Especially the erlang-vm.schema file caused problems when trying to use the NODE_NAME env variable
* Configuration uses the new .src functionality to enable native OS variable replacement in the root configuration file
* Split up smaller configuration files, currently only for the network configuration
* Enabled packing files (data antidote, data riak, logging files) in a root directory if specified via ROOT_DIR_PREFIX
* Same for logging and data dirs (LOGGER_DIR_PREFIX, DATA_DIR_PREFIX)
* Removed vars.config, it was only used for the Mustache templating system of Rebar, which was only used for the schema files of cuttlefish, which we do not use
* Fixed ranch to version 1.7.1 for now, otherwise it would upgrade to a release candidate with breaking API changes
* Bumped antidote to version 0.2.1 (both rebar.config and antidote.app.src)
* Moved documentation from antidote.app.src to config/sys.config where all configuration for the release resides
* Added a validation function to ensure the antidote data directory exists
* Commented out stats calls which are not implemented yet in the antidote_stats module (metadata_updated, metadata_update_stable)
* Removed partially obsolete max ETS tables argument. The amount is not limited anymore, and 8000 is a reasonable default size of the internally used table for named tables.
* Removed the ERL_MAX_PORTS env parameter. This is no longer used and is replaced by vm argument +Q. Furthermore, the default is much higher (65536) than this value (4096).
* Added kernel network config for documentation purposes 
* Removed eunit_opts
    * Verbose output is worse than default output, as it doesn't show failures at the end
    * Eunit_logs is not useful
* Added eunit.coverdata to coverage stats, coverage has increased by ~7%
* Moved monitoring setup to `antidote_stats` repository
* Fixed logging output path in systests and removed copied erlang code for `ct_slave_ext`
* Redirected lager output to logger via custom lager backend

Consequences:

* [x] Update readme for monitoring setup in `antidote_stats` repository
* [x] Update documentation for monitoring setup in Antidote documentation
* [x] Docker files have to be updated
* [x] Update documentation for Antidote on how to use a release, go into more detail than the readme
* Some configuration parameters cannot be directly configured by OS variables and have to be configured dynamically via the `-Application Key Value` parameter tuple, for which `Value` can be used for OS variable configuration
* VM arguments are copy pasted from a rebar default from many years ago. Check if vm args improve performance, if not disable them
* Check if there is a way to specify coverage output path for `ct_run`


albsch:
The new dockerfile can be seen [here](https://github.com/AntidoteDB/docker-antidote/tree/v0.4.0/v0.4.0).

bieniusa:
Why are we bumping from v0.1.1. to v0.4.0?

albsch:
> Why are we bumping from v0.1.1. to v0.4.0?

There are multiple versions everywhere and I didn't know which version to take. The `docker-antidote` version is later than `v0.2.0` and the release configuration (rebar) is `0.0.2`. So I guess the version should be something like `v0.2.1` everywhere? Initially I thought the docker-compose files will stop working, but they are still working.

albsch:
Why is coveralls showing a decrease in coverage? There shouldn't be, does it have problems with it `git rebase`? It seems to compare the coverage to the previous build.

albsch:
New statistics:
image
image
image

Issue #396, opened 2019-10-30
Make antidote_dc_manager:create_dc idempotent


peterzeller:
Running the connect script only works the first time, because of `antidote_dc_manager:create_dc`.

peterzeller:
#399 must be fixed before this can be done.

PR #405, opened 2019-11-20
Use DC manager instead of configuring the test setup manually

albsch:
* Use `antidote_dc_manager` to setup `systests`
* Removed unused code in `test_utils`
* Only create a ring and connect dcs if nodes are started for the first time (can be removed once #396 is fixed)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: erlang
 otp_release:
-  - 21.2
+  - 21.3
+  - 22.1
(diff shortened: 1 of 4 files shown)


albsch:
The `logging_vnode` function `set_sync_log(Value)` is not called anymore after changing to the `antidote_dc_manager` calls.

albsch:
Branch is ready to be merged into `master`.

Issue #400, opened 2019-11-11
Connected DCs metadata not updated


albsch:
Antidote provides a function to get the currently connected data centers via `dc_meta_data_utilities:get_dc_ids/1`.

This is currently used for...

* ... two times for the GentleRain protocol. Currently, the `self` dc returned, resulting in bad snapshot calculations
* ... and for `antidote_stats` to show how many DCs are currently connected

This function only returns the `self` data center, as the `set_dc_partitions` function is never called.

albsch:
A test case was added which fails currently in branch `400-connected-dcs-metadata-not-updated`.

albsch:
Updated test cases. 
`get_dc_ids` is not returning the data centers connected, but instead returns the members of the `riak_core` ring. `get_dc_descriptors` is used to get the connected data centers.
Somewhat unexpectedly, `get_dc_descriptors` returns the value `3` for the system test cluster configuration ([[n1,n2], [n3], [n4]), whereas I would expect each node to know 2 external data centers instead of 3.

Furthermore, what fails now is `shard_count`, which uses the `get_dc_ids` function to get all members. We use our own metadata for that, which is not implemented correctly it seems.
It should be noted that `riak_core` already maintains ring information by itself and the setup is waiting for the ring to converge anyway.

albsch:
The `external_descriptors` name is a bit misleading. It actually means all known (external-facing) descriptors, including self.

PR #408, opened 2019-11-27
Improve DC and InterDC Functionality

### LINKED ISSUE #396: Make antidote_dc_manager:create_dc idempotent

albsch:
Fixes #396 #400 #399.

* Removed redundant stable meta data for our ring. Contrary to the comment, the riak cluster_name does not change after a fail and restart of a node. This fixes a bug where the meta data was not updated correctly and requests were routed to the wrong node, causing handoffs to be triggered.
* Improved documentation of some functions
* Improved adding nodes to a DC 
  * Rewrote DC creation functions
  * `create_dc` is now deprecated but still available
  * `add_nodes_to_dc` is a better fitting function name. Adding nodes to the dc dynamically should be possible, but not tested yet
* Added test cases for #396 and #400 
* Added test case to check for correct shard count per DC
* `add_nodes_to_dc` is now idempotent

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -185,7 +185,9 @@
 -type op() :: {op_name(), op_param()}.
 -type effect() :: term().
 
--type dcid() :: 'undefined' | {atom(),tuple()}. %% TODO, is this the only structure that is returned by riak_core_ring:cluster_name(Ring)?
+
+%% DC Id is the riak_core ring cluster name
+-type dcid() :: term().
 -type snapshot_time() :: 'undefined' | vectorclock:vectorclock().
 -type clock_time() :: non_neg_integer().
 -type dc_and_commit_time() :: {dcid(), clock_time()}.
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -187,7 +187,7 @@
 
 
 %% DC Id is the riak_core ring cluster name
--type dcid() :: term().
+-type dcid() :: undefined | riak_core_ring:riak_core_ring().
 -type snapshot_time() :: 'undefined' | vectorclock:vectorclock().
 -type clock_time() :: non_neg_integer().
 -type dc_and_commit_time() :: {dcid(), clock_time()}.
--- a/src/antidote_dc_manager.erl
+++ b/src/antidote_dc_manager.erl
@@ -45,41 +45,22 @@
 -include("inter_dc_repl.hrl").
 -include_lib("kernel/include/logger.hrl").
 
--export([create_dc/1,
-         get_connection_descriptor/0,
-         subscribe_updates_from/1]
+-export([
+    create_dc/1,
+    get_connection_descriptor/0,
+    subscribe_updates_from/1]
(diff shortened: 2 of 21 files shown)


include/antidote.hrl
```diff
@@ -185,7 +185,9 @@
 -type op() :: {op_name(), op_param()}.
 -type effect() :: term().
 
--type dcid() :: 'undefined' | {atom(),tuple()}. %% TODO, is this the only structure that is returned by riak_core_ring:cluster_name(Ring)?
+
+%% DC Id is the riak_core ring cluster name
+-type dcid() :: term().
```
peterzeller: Was the old type not correct?

src/antidote_dc_manager.erl
```diff
@@ -45,41 +56,23 @@
...
-    %% Ensure each node owns a portion of the ring
-    wait_until_nodes_agree_about_ownership(Nodes),
-    ok = wait_until_no_pending_changes(Nodes),
-    wait_until_ring_converged(Nodes),
-    ok = wait_until(hd(Nodes), fun wait_init:check_ready/1),
-    %% starts metadata services needed for intra-dc communication
-    ok = inter_dc_manager:start_bg_processes(stable_time_functions),
-    ok.
+-spec add_nodes_to_dc([node()]) -> ok | {error, ring_not_ready}.
+add_nodes_to_dc(Nodes) ->
```
peterzeller: We probably should add this function to the PB interface as well?

src/antidote_dc_manager.erl
```diff
@@ -45,41 +56,23 @@
...
+    %% check if ring is ready first
+    case riak_core_ring:ring_ready() of
+        true -> join_new_nodes(Nodes);
+        _ -> {error, ring_not_ready}
```
peterzeller: Why not wait until the ring is ready? 
Or are there cases where it will never be ready after some waiting?

include/antidote.hrl
albsch: This typing definition took a peek into the internals of `riak_core`. We do not know the type, we use it as a transparent `term` for our use cases.

src/antidote_dc_manager.erl
albsch: I don't know how to do that yet, can you add it?

src/antidote_dc_manager.erl
albsch: Ring modifications are rare and should be planned with care. I'd rather throw an error to the client indicating a ring modification of some other user at the same time rather than waiting for the current ring to be ready and getting maybe unexpected ring changes for both users.

It would be no problem to add a sleep and retry the function call, though.

include/antidote.hrl
peterzeller: You can probably change it to:
```suggestion
-type dcid() :: undefined | riak_core_ring:riak_core_ring().
```

PR #411, opened 2019-12-06
Migrate to riak_core_lite

albsch:
This replaces a the `riak_core` version with a minimzed version.

* Removes a large part of unnecessary dependencies introduced by `riak_core` 
* Gets rid of the C-code `eleveldb` dependency.
* Removes lager from the dependency tree


albsch:
This branch is ready to be benchmarked.

PR #414, opened 2020-02-07
cleaning up calls to lists:foldl

shamouda:
Reviewing calls to lists:foldl and optimizing them when possible (e.g. by replacing foldl with a list comprehension or a more efficient list function)

### DIFF
--- a/src/clocksi_vnode.erl
+++ b/src/clocksi_vnode.erl
@@ -154,12 +154,12 @@ send_min_prepared(Partition) ->
 
 %% @doc Sends a prepare request to a Node involved in a tx identified by TxId
 prepare(ListofNodes, TxId) ->
-    lists:foldl(fun({Node, WriteSet}, _Acc) ->
+    lists:foreach(fun({Node, WriteSet}) ->
         riak_core_vnode_master:command(Node,
             {prepare, TxId, WriteSet},
             {fsm, undefined, self()},
             ?CLOCKSI_MASTER)
-    end, ok, ListofNodes).
+    end, ListofNodes).
 
 
 %% @doc Sends prepare+commit to a single partition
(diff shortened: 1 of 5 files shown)


src/clocksi_vnode.erl
```diff
@@ -150,15 +150,7 @@ get_active_txns_internal(TableName) ->
...
-                        lists:foldl(fun({_Key, List}, Acc) ->
-                            case List of
-                                [] ->
-                                    Acc;
-                                _ ->
-                                    List ++ Acc
-                            end
-                        end,
-                            [], [{Key1, List1} | Rest1])
+                        lists:flatmap(fun({_, _List}) -> _List end, [{Key1, List1} | Rest1])
```
bieniusa: Why `_List`? Couldn't it be simply `List`?

src/clocksi_vnode.erl
```diff
@@ -150,15 +150,7 @@ get_active_txns_internal(TableName) ->
                     [] ->
```
albsch: This case check can be left out completely, no?
`lists:flatmap` also works on the empty list `[]`, and we do not do anything special if the list has at least one element.

src/clocksi_vnode.erl
shamouda: Just committed a fix: https://github.com/AntidoteDB/antidote/pull/414/commits/cb6ce9fc54e26929a16250b8fc33a487e5905154

src/clocksi_vnode.erl
shamouda: Just committed a fix: https://github.com/AntidoteDB/antidote/pull/414/commits/cb6ce9fc54e26929a16250b8fc33a487e5905154

PR #415, opened 2020-02-23
antidote_app: Add file comment

### LINKED ISSUE #155: Add file comments for each source module
Each file contains only the copyright info. It would be good to have a 1-5 lines at the top of each file, describing the module.

palash25:
Related to: #155

### DIFF
--- a/src/antidote_app.erl
+++ b/src/antidote_app.erl
@@ -25,6 +25,11 @@
 %% List of the contributors to the development of Antidote: see AUTHORS file.
 %% Description and complete License: see LICENSE file.
 %% -------------------------------------------------------------------
+
+%% @doc Implements application callbacks, starts the antidote supervisor
+%% process, registers riak_core applications and stops the server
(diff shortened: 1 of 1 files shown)


src/antidote_app.erl
```diff
@@ -25,6 +25,11 @@
 %% List of the contributors to the development of Antidote: see AUTHORS file.
 %% Description and complete License: see LICENSE file.
 %% -------------------------------------------------------------------
+
+%% @doc Starts the antidote supervisor process,
```
bieniusa: I suggest adding as information that this module implements the application callbacks.

palash25: would you like that to be added along with the current comment or do you want it to replace it? @bieniusa

bieniusa: Adding :) Thanks!

palash25:
@bieniusa amended please take a look

PR #421, opened 2020-05-14
Improve Interdc communication

dajenet:
This PR limits the entries per request read from log for recovering missing transactions. Entries from log can be read using continuations. Reducing load and prevent possible crashes.
Timeout added for request that read from log.


albsch:
Rebased on top of master.

PR #424, opened 2020-08-28
Refactored bcounter_mgr and corresponding tests

FairPlayer4:
It should now work with key and {key, bucket}
Added TODOs for potential problems or questions

### DIFF
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -43,7 +43,7 @@
 %% This can break the tests is not set to 0
 -define(OLD_SS_MICROSEC,0).
 %% The number of supervisors that are responsible for
-%% supervising transaction coorinator fsms
+%% supervising transaction coordinator fsms
 -define(NUM_SUP, 100).
 %% Threads will sleep for this length when they have to wait
 %% for something that is not ready after which they
@@ -73,11 +73,11 @@
 -define(TRANSFER_FREQ, 100). %in Milliseconds
 
 %% The definition "FIRST_OP" is used by the materializer.
-%% The materialzer caches a tuple for each key containing
+%% The materializer caches a tuple for each key containing
 %% information about the state of operations performed on that key.
 %% The first 3 elements in the tuple are the following meta-data:
 %% First is the key itself
-%% Second is a tuple defined as {the number of update operations stored in the tupe, the size of the tuple}
+%% Second is a tuple defined as {the number of update operations stored in the tuple, the size of the tuple}
 %% Third is a counter that keeps track of how many update operations have been performed on this key
 %% Fourth is where the first update operation is stored
 %% The remaining elements are update operations
@@ -128,7 +128,7 @@
 
 %% The way records are stored in the log.
 -record(log_record, {
-    %% The version of the log record, for backwards compatability
+    %% The version of the log record, for backwards compatibility
     version :: non_neg_integer(),
     op_number :: op_number(),
     bucket_op_number :: op_number(),
--- a/include/antidote.hrl
+++ b/include/antidote.hrl
@@ -83,13 +83,6 @@
 %% The remaining elements are update operations
 -define(FIRST_OP, 4).
 
--record(payload, {
-    key :: key(),
-    type :: type(),
-    op_param,
-    actor :: actor()
-}).
-
 -record(commit_log_payload, {
     commit_time :: dc_and_commit_time(),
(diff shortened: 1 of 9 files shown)


src/dc_meta_data_utilities.erl
```diff
@@ -122,7 +122,8 @@ get_dc_descriptors() ->
         error ->
             ?LOG_DEBUG("Could not read shared meta data for external_descriptors"),
             %% return self descriptor only
-            [inter_dc_manager:get_descriptor()]
+            {ok, Descriptor} = inter_dc_manager:get_descriptor(),
+            [Descriptor]
```
FairPlayer4, line 126: Bugfix (found it during testing single DC)

src/clocksi_downstream.erl
```diff
@@ -35,8 +35,8 @@
 %% @doc Returns downstream operation for upstream operation
 %%      input: Update - upstream operation
 %%      output: Downstream operation or {error, Reason}
--spec generate_downstream_op(Transaction :: tx(), Node :: index_node(), Key :: key(),
-  Type :: type(), Update :: op_param(), list()) ->
+-spec generate_downstream_op(tx(), index_node(), {key(), bucket()},
+    type(), op_param(), list()) ->
```
FairPlayer4, line 39: Makes it clear that key here is always a tuple of key and bucket -> relevant for assumptions in bcounter_mgr

src/logging_vnode.erl
```diff
@@ -779,11 +779,13 @@ handle_commit(TxId, OpPayload, T, Key, MinSnapshotTime, MaxSnapshotTime, Ops, Co
...
-    ((MinSnapshotTime == undefined) orelse (vectorclock:ge(SnapshotTime, MinSnapshotTime))).
+-spec check_min_time(snapshot_time(), snapshot_time() | undefined) -> boolean().
+check_min_time(_SnapshotTime, undefined) -> true;
+check_min_time(SnapshotTime, MinSnapshotTime) -> vectorclock:ge(SnapshotTime, MinSnapshotTime).
 
-check_max_time(SnapshotTime, MaxSnapshotTime) ->
-    ((MaxSnapshotTime == undefined) orelse (vectorclock:le(SnapshotTime, MaxSnapshotTime))).
+-spec check_max_time(snapshot_time(), snapshot_time() | undefined) -> boolean().
+check_max_time(_SnapshotTime, undefined) -> true;
+check_max_time(SnapshotTime, MaxSnapshotTime) -> vectorclock:le(SnapshotTime, MaxSnapshotTime).
```
FairPlayer4, line 788: Slight coverage decrease (one line) but looks a bit nicer and more readable in my opinion

include/antidote.hrl
```diff
@@ -182,10 +175,11 @@
 -type op() :: {op_name(), op_param()}.
 -type effect() :: term().
 
+-type microsecond() :: non_neg_integer().
 
 %% DC Id is the riak_core ring cluster name
--type dcid() :: undefined | riak_core_ring:riak_core_ring().
--type snapshot_time() :: 'undefined' | vectorclock:vectorclock().
+-type dcid() :: undefined | {term(), term()}.
+-type snapshot_time() :: vectorclock:vectorclock().
```
FairPlayer4, line 182: I checked all usages of snapshot_time() 
-> undefined is not allowed except in some places of logging_vnode.erl
-> added undefined at places where it is explicitly allowed (see logging_vnode.erl changes)

src/bcounter_mgr.erl
```diff
@@ -31,201 +31,294 @@
...
+        true ->
+            MyDCID = dc_utilities:get_my_dc_id(),
+            case ToDCID /= MyDCID of
+                true -> antidote_crdt_counter_b:downstream({transfer, {Amount, ToDCID, MyDCID}}, BCounter);
+                false ->
+                    %%TODO this transfer is equal to increment
+                    generate_downstream(_Key, {increment, {Amount, MyDCID}}, BCounter)
+            end;
+        false -> {error, invalid_dcid}
+    end.
```
FairPlayer4, line 104: Here I added some restrictions
-> Transfer only allowed from MyDCID
-> Transfer only allowed to valid DCIDs (descriptor exists)
-> Self-transfer is still allowed but it is made explicit in the code that this is equal to an increment and changes the total permissions of the bcounter

PR #425, opened 2020-09-11
Update and simplify Interdc Communication

albsch:
Some parts of the interdc modules were written with an old architecture of Antidote in mind.
Furthermore, `erlzqm` currently does not compile under Erlang 23 und uses C code, which we would like to avoid.
This rewrite fixes these two problems.

* all modules which open and use a zeromq socket (`pub`, `sub`, `router`, `req`) are named as such
* `inter_dc_query_req` was written with a **socket per partition** architecture in mind, which was changed at some time. Currently, only **one socket per physical node per data center** is used, which allows for simplification of the internal state of `inter_dc_query_req`
* `chumak` is now used instead of `erlzmq`, which makes Antidote almost C-code free. The optional stats dependency is the last dependency with C code (prometheus)
* Improved supervision tree. All of the inter-dc communication is under its own supervisor
* Erlang 23 is now supported
* Bumped `antidote_stats` to the latest version (version bump of transitive dependencies)
* test suite for inter-dc communication with chumak
* eunit tests with mocks for inter dc modules

Some miscellaneous changes:

* Inter-dc publish messages were prefixed with the partition number to be able to use partitions as topic filters. To be able to filter partition number 0 correctly, a binary `<<"P">>` prefix was added to all broadcast messages, which is filtered on receive

* [x] ~~Fix dialyzer errors for `clocksi_coordinator` which are detected by using Erlang 23~~
    * Removed Erl23 from the pipeline and created a separate issue
* [x] ~~Feedback for if we want to use the folder structure for inter dc or keep the module structure flat~~
    * Does not work with the new `rebar3` version, coverage is decreasing when using `rebar3 eunit -d src` as opposed to `rebar3 eunit` which crashed with a non-flat folder structure


src/antidote.erl
```diff
@@ -57,6 +57,7 @@
 
 -spec start() -> {ok, [atom()]} | {error, reason()}.
 start() ->
+    ok = application:ensure_started(chumak),
```
FairPlayer4, line 60: Is this line still necessary? Normally ensure_all_started should also take care of chumak since it is defined under applications in antidote.app.src
Or is it different here?

src/inter_dc/reqrouter/inter_dc_query_req.erl
```diff
@@ -0,0 +1,248 @@
...
+                            {ok, Socket};
+                        _ ->
+                            connect_to_node(Rest)
+                    end;
+                _ ->
+                    connect_to_node(Rest)
+            end;
+        _ ->
+            connect_to_node(Rest)
+    end.
```
FairPlayer4, line 232: Question: Do we have to close the socket and open it again? Couldn't we just return the socket we opened initially and depending on the response close it?

src/inter_dc/reqrouter/inter_dc_query_req.erl
```diff
@@ -0,0 +1,248 @@
...
+            ?LOG_DEBUG("Send and receive"),
+            ok = chumak:send(Socket, FullRequest),
+            {ok, BinaryMsg} = chumak:recv(Socket),
+
+            ?LOG_DEBUG("Process and apply function"),
+            <<_ReqIdBinary:?REQUEST_ID_BYTE_LENGTH/binary, RestMsg/binary>> = inter_dc_utils:check_message_version(BinaryMsg),
+            case RestMsg of
+                <<RequestType, RestBinary/binary>> -> Func(RestBinary);
+                Other -> ?LOG_ERROR("Received unknown reply: ~p", [Other])
+            end,
```
FairPlayer4, line 156: If I understand this correctly, that would be a significant change in the behavior compared to the previous version.
Now the caller would have to wait until a request is completed which is not the right behavior I think.
All processes that send requests (e.g. the bcounter_mgr) would be blocked if the connection is slow or not available.

albsch:
Why is dialyzer suddenly throwing an error for `ranch_protocol`? I didn't touch anything related to that:
```
src/antidote_pb_protocol.erl
  33: Callback info about the ranch_protocol behaviour is not available
```

FairPlayer4:
Dialyzer fixes for OTP 23 (see Slack channel)
grafik

albsch:
The dialyzer fixes will be tracked in another issue ( #426 ).

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+                {ok, _PeerPid} = chumak:connect(Socket, tcp, "localhost", 15556),
+                WorkerLoop(Socket, Id, Parent)
+            end
+        ) end,
+
+
+    StartWorker(),
+    StartWorker(),
+
+    timer:sleep(50),
```
bieniusa, line 393: Is this deterministic and stable? Wouldn't it be better to wait for a specific reply?

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+            BinInt = integer_to_binary(Counter),
+            chumak:send_multipart(RouterSocket, [Id, <<>>, <<BinInt/binary, Msg/binary>>]),
+            L(Counter + 1)
+                    end,
+        spawn_link(fun() -> CaseCheck(0) end),
+        receive finish_test -> ok end,
+        inter_dc_utils:close_socket(RouterSocket)
+          end),
+
+    timer:sleep(50),
```
bieniusa, line 417: See above.

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+
+    %% request identity is strictly required by chumak
+    {ok, Socket} = chumak:socket(req, "A"),
+    {ok, _PeerPid} = chumak:connect(Socket, tcp, "localhost", 15557),
+    WorkerLoop(Socket, self()),
+
+    receive finish -> ok end,
+    RouterPid ! finish_test,
+    inter_dc_utils:close_socket(Socket),
+    timer:sleep(50),
```
bieniusa, line 452: Why is this required?

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+        chumak:subscribe(SubSocket2, P3Bin),
+
+        {ok, _} = chumak:connect(SubSocket2, tcp, "localhost", 15551),
+        timer:sleep(50),
+        Self ! more,
+
+        {ok, _MMM1} = chumak:recv(SubSocket2),
+        {ok, _MMM2} = chumak:recv(SubSocket2),
+
+        %% subscriber_timeout!
```
bieniusa, line 323: Clean up?

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+          end),
+
+    timer:sleep(50),
+
+    % internal API, broadcast allows only for transactions
+    ok = gen_server:call(Pid, {publish, M0}),
+
+    receive more -> ok after 100 -> throw(subscriber_timeout) end,
+
+    ct:log("Publishing ~p",[M1]),
```
bieniusa, line 338: Write as map over list?

test/multidc/inter_dc_zmq_SUITE.erl
```diff
@@ -0,0 +1,453 @@
...
+    timer:sleep(50),
+
+    % internal API, broadcast allows only for transactions
+    ok = gen_server:call(Pid, {publish, Message}),
+
+    gen_server:stop(Pid),
+    timer:sleep(50),
+    ok.
+
+pub_one_sub_two_topics_partitions(_Config) ->
```
bieniusa, line 274: It would be helpful to have a few comments structuring this longer test case.

bieniusa (changes requested):
This was quite a bit of work! I suggest to cleanup just the test cases a little bit.

test/multidc/inter_dc_zmq_SUITE.erl
albsch: The subscriber_timeout comment is just a documentation on when the timeout on the chumak side happens, if you uncomment that line of code

test/multidc/inter_dc_zmq_SUITE.erl
albsch: It's a bit hard to get rid of all sleeps with sockets, but I tried to change most of the sleeps to message passing.

bieniusa (approved):

PR #428, opened 2020-09-13
Dialyzer fixes for OTP 23

FairPlayer4:
The first execute_op must be deleted because it will never be called (there is no defined timeout so it can never happen)
The From/Sender is defined as either undefined or gen_statem:from() (which is {pid(), term()}) so it can never be just a pid


FairPlayer4:
The failed build for OTP 21.3 is interesting

PR #439, opened 2020-09-15
Refactored bcounter_mgr and corresponding tests

FairPlayer4:
Probably needs review first

### DIFF
--- a/src/bcounter_mgr.erl
+++ b/src/bcounter_mgr.erl
@@ -365,6 +365,7 @@ can_transfer(Key, _, RequesterDCID, LastTransfers) ->
 -ifdef(TEST).
 
 %%TODO use meck
+
 bcounter_mgr_test() ->
     {foreach, fun setup/0, fun cleanup/1, [
         fun server_is_alive/1
--- a/src/bcounter_mgr.erl
+++ b/src/bcounter_mgr.erl
@@ -36,6 +36,13 @@
 %% TODO check whether only local operations are allowed (currently yes to prevent odd behaviour)
 
 -module(bcounter_mgr).
+
+-ifdef(TEST).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-endif.
+
 -behaviour(gen_server).
 
 -export([generate_downstream/3,
@@ -357,9 +364,11 @@ can_transfer(Key, _, RequesterDCID, LastTransfers) ->
 
 -ifdef(TEST).
 
--include_lib("eunit/include/eunit.hrl").
-
 %%TODO use meck
+bcounter_mgr_test() ->
+    {foreach, fun setup/0, fun cleanup/1, [
+        fun server_is_alive/1
+    ]}.
 
 setup() ->
     process_flag(trap_exit, true),
@@ -370,11 +379,6 @@ cleanup(Pid) ->
     exit(Pid, normal),
     timer:sleep(50), %%TODO maybe check with recursion and lower intervals
     ?assertEqual(false, is_process_alive(Pid)).
-
-bcounter_mgr_test() ->
-    {foreach, fun setup/0, fun cleanup/1, [
-        fun server_is_alive/1
-    ]}.
 
 server_is_alive(Pid) ->
     fun() ->
--- a/src/bcounter_mgr.erl
+++ b/src/bcounter_mgr.erl
@@ -31,201 +31,294 @@
 %% Basic inter-dc reservations manager only requests remote reservations
 %% when necessary.
 %% Transfer requests are throttled to prevent distribution unbalancing
+%% Can be crashed and restarted without too much harm (only the temporary state is lost)
 %% (TODO: implement inter-dc transference policy E.g, round-robin).
+%% TODO check whether only local operations are allowed (currently yes to prevent odd behaviour)
 
 -module(bcounter_mgr).
 -behaviour(gen_server).
 
--export([start_link/0,
-         generate_downstream/3,
-         process_transfer/1,
-         request_response/1
-        ]).
+-export([generate_downstream/3,
+    process_transfer/1,
+    request_response/1,
+    get_pending_transfer_requests/0,
+    set_transfer_periodic_active/1]).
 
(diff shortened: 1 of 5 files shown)


FairPlayer4:
- Transfers were To and From are equal (called self transfers) are always allowed and translated to an increment
- The CRDT antidote_crdt_counter_b prohibits self transfers if no local increment exists and if one exists then self transfers are essentially equal to increments
- The new version is a bit more restrictive on transfers and allows only transfers that go to an existing DC (a descriptor exists)
- (Like the previous version) Increments and decrements are only allowed locally and the target DC that is defined in the update operation is ignored

marc-shapiro:
> Le 15 sept. 2020 à 16h56, Kevin Bartik <notifications@github.com> a écrit :
> 
> (Like the previous version) Increments and decrements are only allowed locally and the target DC that is defined in the update operation is ignored

Wouldn’t it be better to raise an error if target DC ≠ local DC?

					Marc

FairPlayer4:
I can add a check that returns {error, no_permissions} if an update operation contains an invalid DC
It forces the client to make the right call which has advantages (always intended behavior) and disadvantages (client must find out the DCID when they want to perform an operation on a bcounter)

bieniusa (approved):

src/bcounter_mgr.erl
```diff
@@ -31,201 +31,294 @@
...
 
-handle_info(transfer_periodic, #state{req_queue=RQ0, transfer_timer=OldTimer}=State) ->
+handle_info(transfer_periodic, State = #state{transfer_periodic_active = false}) -> {noreply, restart_timer(State)};
+handle_info(transfer_periodic, State = #state{pending_transfer_requests = PendingTransferRequests}) ->
+    NewPendingTransferRequests = transfer_periodic(PendingTransferRequests),
+    {noreply, restart_timer(State#state{pending_transfer_requests = NewPendingTransferRequests})}.
+
+terminate(_Reason, #state{transfer_timer = OldTimer}) ->
     _ = erlang:cancel_timer(OldTimer),
-    RQ = clear_pending_req(RQ0, ?REQUEST_TIMEOUT),
```
bieniusa, line 130: Why is this simplification valid?

test/multidc/bcountermgr_SUITE.erl
```diff
@@ -30,25 +30,24 @@
...
+    conditional_write_test_run/1,
+    parallel_increment_decrement/1,
+    two_nodes_want_everything/1]).
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
--define(TYPE, antidote_crdt_counter_b).
--define(BUCKET, test_utils:bucket(bcountermgr_bucket)).
+-define(BUCKET, test_utils:bucket(bcountermgr_bucket)). %%TODO Why?
```
bieniusa, line 50: To ensure that different test cases don't interfere. Resetting the test fixture is too time consuming.

bieniusa (changes requested):

src/bcounter_mgr.erl
FairPlayer4: It was replaced by 
`ClearedPendingTransferRequests = clear_old_transfer_requests(PendingTransferRequests, ?REQUEST_TIMEOUT),`
in line 214
I moved the code a bit so that the gen_server callbacks don't contain the logic (this can be reverted)

test/multidc/bcountermgr_SUITE.erl
FairPlayer4: Ok I will remove those TODOs

marc-shapiro:
Best of both worlds:
by default the system fills in the correct DCID when none is provided
When one is provided, check that it is correct

					Marc

bieniusa:
I am still a bit confused why a local *transfer* would result in an increment. What is the semantic reasoning behind this?

FairPlayer4:
> 
> 
> I am still a bit confused why a local _transfer_ would result in an increment. What is the semantic reasoning behind this?

This can be seen by the internal implementation of the antidote_crdt_counter_b

`-spec increment(id(), pos_integer(), antidote_crdt_counter_b()) -> {ok, antidote_crdt_counter_b()}.
increment(Id, V, {P, D}) ->
    {ok, {orddict:update_counter({Id, Id}, V, P), D}}.`

`-spec transfer(id(), id(), pos_integer(), antidote_crdt_counter_b()) -> {ok, antidote_crdt_counter_b()}.
transfer(FromId, ToId, V, {P, D}) ->
    {ok, {orddict:update_counter({FromId, ToId}, V, P), D}}.`

In case of a transfer the available permissions are always checked beforehand and if not enough permissions exist then the operation fails with {error, no_permissions}
The behavior with self transfer is inconsistent because it will add permissions (like increment) only if enough permissions exist in the first place.
In a normal transfer operation no permissions are added, they are only shifted. The self transfer (like the increment) actually increases the globally available permissions.

FairPlayer4:
> 
> Best of both worlds: by default the system fills in the correct DCID when none is provided When one is provided, check that it is correct Marc
>
I would like to do this but this requires further changes (probably in the antidote_crdt repo) because antidote checks the update operation that is provided by the client. Only operations that pass this check actually arrive at the bcounter_mgr. Internally the antidote_crdt_counter_b requires an id() (we always use a DCID) and the is_operation() check requires that such an id() is provided.
Overall the client must provide something for the id() field and currently, it is ignored and always replaced by the local DCID (in case of transfers the FromID is always the local DCID, and the ToID is defined by the client (it is checked and only a valid DCID will pass))

FairPlayer4:
@marc-shapiro 
I will create a solution that fits your requirements (full support will require changes to antidote_crdt_counter_b => allow operations without an id() but that can be done later)
For now, the client can set the DCID to undefined which will use the local DCID
Otherwise, if the client uses an invalid DCID an error will be returned

FairPlayer4:
@bieniusa @marc-shapiro
I will create another pull request that resolves the conflicts with master and runs the CI

PR #442, opened 2020-09-18
Tests for join after replicate

albsch:
Added system test case for joining a cluster to replicate to one more empty DC. Tests that requests for missing log entries are requested properly by `inter_dc_query_dealer`.
Added more time for socket to open in `inter_dc_zmq_SUITE`.


albsch:
Closing for now, there is still some timing bug left.

PR #446, opened 2020-09-22
Use the antidote_crdt module instead of the CRDT type

mrshankly:
With the introduction of the new CRDT type aliases in AntidoteDB/antidote_crdt#40, it is now possible that some CRDTs do not have an actual implementation server side, they are simply mapped to an implementation of another CRDT.

Using `Type:function(...)` instead of `antidote_crdt:function(Type, ...)` results in an undefined error when `Type` is an alias. This PR fixes that.


FairPlayer4 (approved):
Looks good and makes sense

PR #451, opened 2021-05-19
Testing Refactor

albsch:
* Simplified testing setup and starting nodes in `ct` utils
* Rewrote clocksi mock to use `meck` instead of macros for mocking
* Fixed a dialyzer issue
* Updated `rebar3` executable
* Fixed some new Erlang 24 compile errors
* Github workflows instead of travis
* Added Erlang 24 support


.github/workflows/erlang.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+    strategy: 
+      fail-fast: false
+      matrix:
+        erlang: [ 21,22,23 ]
```
bieniusa, line 15: Is there already support for 24.0? We can probably drop 21.

bieniusa (approved):
Very nice cleanup!

.github/workflows/erlang.yml
albsch: Yes, the workflow passes for Erlang 24, too. I added it.

PR #459, opened 2021-12-30
Add a read or validate API

liaud:
This is the first steps for a read or validate API. I plan to update protobuf files when we will have settled on this.

The API takes a arbitrary token as an input, if the token is valid
the client can use the cached value, otherwise the new value is returned
alongside a token for the next call.

* A token is the commit time of a snapshot, if commit time is different
  a new value is returned. An empty token is a niche value that is
  always invalid.

* When the commit time for a snapshot is unknown, the token is considered
  invalid.

* The materializer has been modified a bit to return the commit time alongside
  the snapshot.

* The materializer is the one who does the check to avoid unnecessary data
  round-trips.

* Transactions with local updates always invalidate the token if updates
  concerns the same partition. This use case is considered rare so no
  optimization have been tried on this.

I was not 100% sure if what I did is fully correct, basics test are working and previous tests still pass, happy to hear feedback on this !

### DIFF
--- a/src/clocksi_interactive_coord.erl
+++ b/src/clocksi_interactive_coord.erl
@@ -63,6 +63,7 @@
     receive_committed/3,
     receive_logging_responses/3,
     receive_read_objects_result/3,
+    receive_validate_or_read_objects_result/3,
     receive_aborted/3,
     single_committing/3,
     receive_prepared/3,
@@ -365,6 +366,58 @@ receive_read_objects_result(info, {_EventType, EventValue}, State) ->
     receive_read_objects_result(cast, EventValue, State).
 
 
+receive_validate_or_read_objects_result(cast, {ok, {invalid, Key, Type, ReadSnapshot, Token}}, CoordState = #state{
+    num_to_read = NumToRead,
+    return_accumulator = ReadKeys,
+    updated_partitions = UpdatedPartitions
+}) ->
+    {Snapshot, EffectiveToken} = case length(UpdatedPartitions) > 0 of
+        true ->
+            % There are some local updates that must be applied.
+            % It is not safe to apply the update and to return
+            % the token returned by the vnode noor it is safe to return
+            % the updated state with a token specific to this uncommited
+            % transaction.
+            %
+            % In such case, always return a token that will be invalid so
+            % the client can know that returned value shouldn't be cached for
+            % later revalidation.
+            UpdatedSnapshot = apply_tx_updates_to_snapshot(Key, CoordState, Type, ReadSnapshot),
+            {UpdatedSnapshot, ?INVALID_OBJECT_TOKEN};
+        false ->
+            {ReadSnapshot, Token}
+    end,
+
+    ReadValues = replace_first(ReadKeys, Key, {invalid, Snapshot, EffectiveToken}),
+
+    case NumToRead > 1 of
+        true ->
+            {next_state, receive_validate_or_read_objects_result, CoordState#state{
+                num_to_read = NumToRead - 1,
+                return_accumulator = ReadValues
+            }};
+        false ->
+            {next_state, execute_op, CoordState#state{num_to_read = 0},
+                [{reply, CoordState#state.from, {ok, lists:reverse(ReadValues)}}]}
+    end;
+
+receive_validate_or_read_objects_result(cast, {ok, {valid, Key, _Type}}, CoordState = #state{
+    num_to_read = NumToRead,
+    return_accumulator = ReadKeys
+}) ->
+    ReadValues = replace_first(ReadKeys, Key, valid),
+
+    case NumToRead > 1 of
+        true ->
+            {next_state, receive_validate_or_read_objects_result, CoordState#state{
+                num_to_read = NumToRead - 1,
+                return_accumulator = ReadValues
+            }};
+        false ->
+            {next_state, execute_op, CoordState#state{num_to_read = 0},
+                [{reply, CoordState#state.from, {ok, lists:reverse(ReadValues)}}]}
+    end.
+
 %%%== receive_logging_responses
 
 %% internal state timeout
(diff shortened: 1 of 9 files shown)


test/singledc/antidote_SUITE.erl
```diff
@@ -212,3 +218,138 @@ txn_seq_update_check(Node, TxId, Updates) ->
...
+    Node = proplists:get_value(node, Config),
+    Type = antidote_crdt_counter_pn,
+
+    Object1 = {vor_key1, Type, Bucket},
+
+    {ok, TxIdA1} = rpc:call(Node, antidote, start_transaction, [ignore, []]),
+    {ok, TxIdB1} = rpc:call(Node, antidote, start_transaction, [ignore, []]),
+
+    % On the first read of an empty object without a min transaction time, there is no
+    % other choice but to return an invalid token.
```
liaud, line 273: I am not sure of why this happens and if I can / should prevent it. If someone has some input on this...

src/clocksi_readitem.erl
```diff
@@ -68,6 +71,23 @@ async_read_data_item({Partition, Node}, Key, Type, Transaction, PropertyList, {f
     } end),
     ok.
 
+-spec validate_or_read_data_item(index_node(), key(), type(), object_token(), tx(), read_property_list()) -> {ok, {invalid, snapshot(), object_token()}} | {ok, valid} | {error, term()}.
+validate_or_read_data_item({Partition, Node}, Key, Type, Token, Transaction, PropertyList) ->
```
liaud, line 75: We might want to remove this, it is only used for mock tests in the `interactive_coord_clocksi`. I wanted to mimick what exists for the read API.

bieniusa:
Thank you! There seem to be some issues with the tests, can you recheck?

liaud:
@bieniusa It should be ok now. I assumed that the objects will always lend in the same partition with the given keys, but it seems that it is not the case. So the test was failing time to time.

I tried to use `log_utilities:get_key_partition({voy_key1, Bucket}),` in tests to have a precise assertion but couldn't make it work... If you have an idea on why fails and how to fix it I could improve the assertion:
```
    [{ets,lookup,
         [ets_riak_core_ring_manager,{bucket,<<"antidote">>}],
         [{error_info,#{cause => id,module => erl_stdlib_errors}}]},
     {riak_core_ring_manager,get_bucket_meta,1,
         [{file,
              "/home/romain/devenv/antidote/antidote/_build/default/lib/riak_core/src/riak_core_ring_manager.erl"},
          {line,176}]},
     {riak_core_bucket,get_bucket,1,
         [{file,
              "/home/romain/devenv/antidote/antidote/_build/default/lib/riak_core/src/riak_core_bucket.erl"},
          {line,118}]},
     {riak_core_util,chash_key,1,
         [{file,
              "/home/romain/devenv/antidote/antidote/_build/default/lib/riak_core/src/riak_core_util.erl"},
          {line,235}]},
     {log_utilities,convert_key,1,
         [{file,"/home/romain/devenv/antidote/antidote/src/log_utilities.erl"},
          {line,120}]},
     {log_utilities,get_preflist_from_key,1,
         [{file,"/home/romain/devenv/antidote/antidote/src/log_utilities.erl"},
          {line,70}]},
     {log_utilities,get_key_partition,1,
         [{file,"/home/romain/devenv/antidote/antidote/src/log_utilities.erl"},
          {line,63}]},
     {antidote_SUITE,interactive_txn_validate_or_read_multiple,1,
         [{file,
              "/home/romain/devenv/antidote/antidote/test/singledc/antidote_SUITE.erl"},
          {line,340}]}]}
```

src/clocksi_interactive_coord.erl
```diff
@@ -365,6 +366,58 @@ receive_read_objects_result(info, {_EventType, EventValue}, State) ->
...
+
+    case NumToRead > 1 of
+        true ->
+            {next_state, receive_validate_or_read_objects_result, CoordState#state{
+                num_to_read = NumToRead - 1,
+                return_accumulator = ReadValues
+            }};
+        false ->
+            {next_state, execute_op, CoordState#state{num_to_read = 0},
+                [{reply, CoordState#state.from, {ok, lists:reverse(ReadValues)}}]}
```
bieniusa, line 401: What is the assumption regarding the order here?

src/clocksi_interactive_coord.erl
```diff
@@ -564,6 +617,49 @@ execute_command(read_objects, Objects, Sender, State = #state{transaction=Transa
 
     {next_state, receive_read_objects_result, NewCoordState#state{from = Sender}};
 
+execute_command(validate_or_read, {Key, Type, Token}, Sender, State = #state{
+    transaction=Transaction,
```
bieniusa, line 621: Would probably be more efficient to encode this as filter or fold over the `ReadResult` list.

src/clocksi_interactive_coord.erl
bieniusa: Ah, I see that you are zipping two lists later.

src/clocksi_interactive_coord.erl
```diff
@@ -744,6 +846,22 @@ perform_read({Key, Type}, UpdatedPartitions, Transaction, Sender) ->
             {error, Reason}
     end.
 
+perform_validate_or_read({Key, Type, Token}, UpdatedPartitions, Transaction, Sender) ->
+    Partition = log_utilities:get_key_partition(Key),
+
+    WriteSet = case lists:keyfind(Partition, 1, UpdatedPartitions) of
+        false -> [];
+        {Partition, WS} ->
+            ?INVALID_OBJECT_TOKEN = Token,
```
bieniusa, line 855: Can this check fail? Seems dangerous to encode an assert like this.

src/clocksi_readitem.erl
bieniusa: You could move it into the TEST part, then it is only available when compiled for testing. But I am ok with keeping it similar to the read API.

src/clocksi_vnode.erl
```diff
@@ -110,6 +112,25 @@ read_data_item(Node, TxId, Key, Type, Updates) ->
...
+    case clocksi_readitem:validate_or_read_data_item(Node, Key, Type, Token, TxId, []) of
+        {ok, {invalid, Snapshot, _}} ->
+            Updates2 = reverse_and_filter_updates_per_key(Updates, Key),
+            Snapshot2 = clocksi_materializer:materialize_eager(Type, Snapshot, Updates2),
+
+            {ok, {invalid, Snapshot2, ?INVALID_OBJECT_TOKEN}};
+        {ok, valid} ->
+            % When valid, there shouldn't be any update to apply here and the
+            % provided Token is necessarily invalid.
+            ?INVALID_OBJECT_TOKEN = Token,
```
bieniusa, line 125: See comment above: This check might fail?

src/materializer_vnode.erl
```diff
@@ -444,18 +488,19 @@ fetch_updates_from_cache(OpsCache, Key) ->
...
 -spec materialize_snapshot(txid() | ignore, key(), type(), snapshot_time(), boolean(), state(), snapshot_get_response())
-    -> {ok, snapshot_time()} | {error, reason()}.
+    -> {ok, {snapshot_time(), snapshot()}} | {error, reason()}.
 
-materialize_snapshot(_TxId, _Key, _Type, _SnapshotTime, _ShouldGC, _State, #snapshot_get_response{
+materialize_snapshot(_TxId, _Key, _Type, _MinSnapshotTime, _ShouldGC, _State, #snapshot_get_response{
             number_of_ops=0,
-            materialized_snapshot=Snapshot
+            materialized_snapshot=Snapshot,
+            snapshot_time=CommitTime
```
bieniusa, line 496: I would suggest renaming `CommitTime` here, because a snapshot corresponds to a vector clock time, and this time might *not* correspond to a commit time of a transaction. Consider e.g. the case where a snapshot consists of changes resulting from two concurrent txns. What would be the commit time in this case if both transaction's commit vectors are incomparable?

test/singledc/antidote_SUITE.erl
bieniusa: Can you add some comment to the definition of invalid token explaining its meaning?

bieniusa (approved):

src/clocksi_interactive_coord.erl
liaud: I don't expect this check to fail.

If it fails it means that we might return something incorrect and that there is a programming error on the direct caller.

test/singledc/antidote_SUITE.erl
liaud: Done!

src/clocksi_vnode.erl
liaud: Removed this one. But the other one is fine I believe.

src/materializer_vnode.erl
liaud: Indeed, I chose a name more close to the actual naming convention here and named it `SnapshotTime` hope it is correct ?

I prefer to not introduce the term UniqueToken here and keep this in the validate_or_read API.

liaud:
@bieniusa I think I addressed the main comments and upgraded this PR from the draft to something merge-able. I will come back on the protobuf part and if we think that the user facing API should change, I will redo a PR for the API corrections here.

bieniusa (approved):

bieniusa:
@liaud I had wanted to merge, but @albsch and I keep getting timeouts in some unit tests. We will try to clarify AFAP.

liaud:
@bieniusa Any news on the timeouts ?

bieniusa:
Can you pull master in?

PR #466, opened 2022-03-22
https in dep url instead of git

define-null:
Just a bit more user-friendly experience.


bieniusa:
Thanks for your help! Quick question: Did you manually change the url or were there updates on rebar3?

define-null:
Changed it manually

bieniusa:
I see, we should actually also change it upstream here: https://github.com/AntidoteDB/antidote_stats/blob/master/rebar.config

bieniusa (approved):

PR #468, opened 2022-05-02
merge gingko feature

dch:


albsch (changes requested):
Some comments before I can continue the review:

* The `_checkouts` folder should not be in the main repository, it's already ignored, but not yet removed.
* `gingko` is not yet on `hex.pm` and Antidote cannot be compiled.

When I can compile the project I'll continue the review. Also, at first glance, I can't see why the Github Actions pipeline is not triggered on this pull request. I'll have to double-check why.

A short summary of what the expected changes of this pull request are would be nice. I know what `gingko` is about, but something to review against is always helpful.

dch:
hi @albsch sorry I intended to send this PR to the vaxine repo before it's merged - it needs further work! I'll close this & re-open it when we're ready.

PR #471, opened 2022-05-10
Option 1: Mono Repository with symlinks

albsch:
Included local dependencies `vectorclock`, `antidote-erlang-client`, `antidote_crdt`, `antidote_pb_codec`, `antidote_stats`.

**+**
* No structural changes to any project
* Github workflow working directory change is enough to get workflows running
* Simple

**-**
* No shared configuration (common linting rules, dialyzer rules, etc.), configuration duplication
* Including `_checkouts` feels like a hack
* Explicit local dependency management via symlinks and `rebar3.config` dependencies
* Including the Erlang client in another project as a dependency might be difficult

Other comments:
* `_build` folders are not shared
* Commands are executed in their respective folders
* Local dependencies are not locked to a version


albsch:
Closing in favor of #472

PR #472, opened 2022-05-10
Option 2: Mono Repository with Umbrella Apps Structure

albsch:
Included local dependencies `vectorclock`, `antidote-erlang-client`, `antidote_crdt`, `antidote_pb_codec`, `antidote_stats`.

**+**
* Shared top-level configuration (common linting rules, dialyzer rules, etc.)
* Implicit local code dependency management

**-**
* Duplication of configurations in `rebar3.config`
  * e.g. `C` compiling directives need to be in the top-level and antidote `rebar.config`, if one wants to compile antidote in the stand-alone app directory

Other comments:
* single shared `_build` folder
* Commands are executed at the top-level
* Local dependencies are not locked to a version

Changes:
* bumped `rebar3` version
* default vm.args until benchmarks prove otherwise
* uniform linting for umbrella project
* fixed linting errors
* updated `rebar3_lint` plugin
* unified test coverage
* erlang client removed from antidote dependencies
* erlang client can be included as a standalone dependency in other projects
(path dependency)
* single format run with `erlfmt`
    * fixed comments which were moved one scope outside with replace 
    * From: `^(\s*)%(.*\s*)\n(\s*)\{\n` 
    * To: `$3\{\n    $1% $2\n`
* added `erlfmt --check` to workflow
* added git history of local apps as unrelated merged git histories
* [x] `rebar3 dialyzer` reports two errors:

  apps/antidote/src/clocksi_readitem.erl
  Line 86 Column 13: The pattern {'error', Reason} can never match the type {'ok','valid' | {'invalid',_,binary()}}
  
  apps/antidote/src/materializer_vnode.erl
  Line 387 Column 17: The pattern 'ignore' can never match the type #{_=>pos_integer()}

Breaking changes:
* `antidote_crdt`
    * `lint`: `antidote_crdt_counter_b:localPermissions/2` renamed to `antidote_crdt_counter_b:local-permissions/2`
		* Affects `antidote` `bcounter_mgr`

Todos future (create issues):

* [x] `antidote:inter_dc_query_router:request_log_entries/0` creates
`inter_dc_sub_vnode.(\d+).coverdata` temporary file which is not cleaned up (likely
bug in meck and eunit cover interaction). There are some some warnings showing
up in `cover.log`. (#474)
* [x] `antidote_crdt` test linting (#475)
* [x] `rebar3 release` generates a warning for prometheus source code missing,
but it is included in the release and the release is working fine  (#476)

Todos for this PR:
* [x] License files: Top-level? Per app? Merge?
* [x] `antidote_stats` `rebar3 lint` fails
		* Wait until https://github.com/project-fifo/rebar3_lint/issues/46

Checked (top-level)
* [x] `rebar3 deps`
* [x] `rebar3 eunit` (+coverage)
* [x] `rebar3 proper` (+coverage)
* [x] `rebar3 lint`
* [x] `rebar3 dialyzer`
* [x] `rebar3 xref`
* [x] `rebar3 release`
* [x] `reltest`
* [x] `ct` system tests
* [x] Makefile
* [x] Github workflows
* [x] local Docker image
* [x] local stats endpoint producing data (`localhost:3001/metrics`)
* [x] Prometheus and Grafana docker setup
* [x] Test erlang client in other project
* [x] Code coverage


define-null:
So if I fetch this branch run `rebar3 compile` all works fine, but not for the case when I'd like just to fetch dependencies:
```
$ rebar3 deps
...
===> Fetching quantile_estimator v0.2.1
===> Failed to update package antidote_stats from repo hexpm
===> Package not found in any repo: antidote_stats
```
And that's because `antidote_stats` is mentioned as a dependency in `antidote`, but there is no information how to fetch such a dependency.(it expects hex by default when no information is provided) If I drop `antidote_stats` from dependency list `rebar3 deps` works ok.

As far as I understand the reason why we might want to have local app `antidote_stats` or similar to be listed as a dependency in `app/antidote/rebar.config` is for cases when somebody might want to use `get_subdirs` functionality of the `rebar3`. (Which I would expect for apps like `erlang-antidote-client`).

I haven't personally tested which version we end up with in the `_build` directory when listing `antidote_pb_codec` with `get_subdir` setting in `apps/erlang-antidote-client/rebar.config`, but it would be interesting to see and probably worth checking it. The question here - would we end up with fetching concrete version mentioned in get_subdir, or use local application instead?

albsch:
> The question here - would we end up with fetching concrete version mentioned in get_subdir, or use local application instead?

The local commands (`rebar3 compile`, `rebar3 release`, etc.) will fetch the local versions.

define-null:
Aha, ok.. so if `rebar3 compile` and `rebar3 release` use local version and ignore `git_subdir` dependency, then one solution would be to use `git_subdir` to specify all local deps in `apps/antidote-erlang-client/rebar.config`.

That way when somebody uses `git_subdir` in a separate project to fetch `antidote-erlang-client` it would be also possible to fetch `antidote_pb_codec` dep even if it's not on hex yet.

albsch:
I tried it just now, for some reason `rebar3` doesn't resolve the nested `git_subdir` `antidote_pb_codec` dependency inside the `git_subdir` `antidotec_pb` dependency. I think I read somewhere that `rebar3` doesn't yet support nested umbrella app dependencies. I'll experiment a bit more, but I think it's not supported yet.

albsch:
It works quite well with the `path dep` plugin, I added it. I tried it, including the client in another app works fine now.

bieniusa:
So, it seems that this Option 2 is the way to go!

albsch:
So it seems these two Dialyzer errors are Dialyzer bugs.

apps/antidote/src/clocksi_readitem.erl
```diff
@@ -82,9 +82,10 @@ async_validate_or_read_data_item({Partition, Node}, Key, Type, Token, Transactio
             {ok, {invalid, Snapshot, NextToken}} ->
                 gen_statem:cast(Sender, {ok, {invalid, Key, Type, Snapshot, NextToken}});
             {ok, valid} ->
-                gen_statem:cast(Sender, {ok, {valid, Key, Type}});
-            {error, Reason} ->
-                gen_statem:cast(Sender, {error, Reason})
+                gen_statem:cast(Sender, {ok, {valid, Key, Type}})
```
albsch, line 85: Unsure why dialyzer sees that `{error, Reason}` can never be returned here.

define-null:
Based on commit 7d772b1911d45e3abff2c8776b9d1ee77221d8f5:
This fix eliminates dialyzer warnings for me. Spec correction to take into account `ignore`, but also `{error, no_snapshot}` is not correct.
```
diff --git a/apps/antidote/src/materializer_vnode.erl b/apps/antidote/src/materializer_vnode.erl
index 24c5dcce..21f88a21 100644
--- a/apps/antidote/src/materializer_vnode.erl
+++ b/apps/antidote/src/materializer_vnode.erl
@@ -377,7 +377,7 @@ internal_read(Key, Type, MinSnapshotTime, TxId, _PropertyList, ShouldGc, State)
                                 clocksi_readitem:read_property_list(),
                                 boolean(), state()) -> {ok, valid}
                                                         | {ok, {invalid, snapshot(), object_token()}}
-                                                        | {error, no_snapshot}.
+                                                        | {error, term()}.
 internal_validate_or_read(Key, Type, Token, MinSnapshotTime, TxId, _PropertyList, ShouldGc, State) ->
     SnapshotGetResp = get_from_snapshot_cache(TxId, Key, Type, MinSnapshotTime, State),

@@ -488,7 +488,7 @@ fetch_updates_from_cache(OpsCache, Key) ->
     end.

 -spec materialize_snapshot(txid() | ignore, key(), type(), snapshot_time(), boolean(), state(), snapshot_get_response())
-    -> {ok, {snapshot_time(), snapshot()}} | {error, reason()}.
+    -> {ok, {snapshot_time() | ignore, snapshot()}} | {error, term()}.

 materialize_snapshot(_TxId, _Key, _Type, _MinSnapshotTime, _ShouldGC, _State, #snapshot_get_response{
             number_of_ops=0,
```

define-null:
@albsch @bieniusa 
Taking into account that we import a lot of "new code" here there are 2 suggestions:
- consider full history import instead of the single version, so that we notice if we introduce any accidental changes to the original code
- add necessary format corrections, to fix indentation, long lines, tab/space issues, so that to avoid doing that work in the future.

bieniusa:
Regarding the formatter:
- Using the `rebar3_format` plugin with the `otp_formatter`, it adds quotes to the macros in the binary types in `inter_dc_utils` and breaks compilations. 
-  The `default` version works :)

bieniusa (approved):

albsch:
Thanks @bieniusa and @define-null for helping! Going to merge the PR now.

PR #477, opened 2022-05-28
antidote-crdt: allow multiple operations in the same map key

v0idpwn:
Some CRDT types have operations that can't be composed. For example,
`antidote_crdt_set_rw` operations `add_all` and `delete_all` aren't
composable. This change allows  one to perform both operations in
a single map update.

### DIFF
--- a/apps/antidote_crdt/src/antidote_crdt_map_go.erl
+++ b/apps/antidote_crdt/src/antidote_crdt_map_go.erl
@@ -148,16 +148,12 @@ is_operation(Operation) ->
         {update, {{_Key, Type}, Op}} ->
             antidote_crdt:is_type(Type) andalso antidote_crdt:is_operation(Type, Op);
         {update, Ops} when is_list(Ops) ->
-            distinct([Key || {Key, _} <- Ops]) andalso
-                lists:all(fun(Op) -> is_operation({update, Op}) end, Ops);
+            lists:all(fun(Op) -> is_operation({update, Op}) end, Ops);
         {reset, {}} ->
             false;
         _ ->
             false
     end.
-
-distinct([]) -> true;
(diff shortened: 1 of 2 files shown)


apps/antidote_crdt/src/antidote_crdt_map_go.erl
```diff
@@ -148,8 +148,7 @@ is_operation(Operation) ->
         {update, {{_Key, Type}, Op}} ->
             antidote_crdt:is_type(Type) andalso antidote_crdt:is_operation(Type, Op);
         {update, Ops} when is_list(Ops) ->
-            distinct([Key || {Key, _} <- Ops]) andalso
```
bieniusa, line 151: Dialyzer is saying that `distinct\1` is now not used any more. I suggest to completely remove it.

apps/antidote_crdt/src/antidote_crdt_map_rr.erl
```diff
@@ -223,8 +223,7 @@ is_operation(Operation) ->
         {update, {{_Key, Type}, Op}} ->
             antidote_crdt:is_type(Type) andalso antidote_crdt:is_operation(Type, Op);
         {update, Ops} when is_list(Ops) ->
-            distinct([Key || {Key, _} <- Ops]) andalso
-                lists:all(fun(Op) -> is_operation({update, Op}) end, Ops);
+            lists:all(fun(Op) -> is_operation({update, Op}) end, Ops);
```
bieniusa, line 226: There are a few more calls for `distinct` check further down, I think they can (and should) also be removed.

bieniusa (changes requested):
Thank you for the fast PR!

bieniusa (approved):

PR #479, opened 2022-05-30
antidote-crdt: add tests for multi ops on same key (map types)

### LINKED ISSUE #478: Add test cases for multiple updates on the same key
In #477 we relaxed the requirements on squashing updates on the same CRDT in a single transactions into one operation as this is infeasible for some operations and CRDT types.
Need to add a test to document that this is allowed now.

v0idpwn:
Closes #478


bieniusa:
Can you add a final one for the sets?

PR #483, opened 2022-06-08
Rename antidote-erlang-client dir to conform with erlang application name

define-null:


bieniusa:
Can you pull latest master to ensure that the CI passes now?

bieniusa:
Also, it would be helpful to have a commit message with a bit more information ;)

define-null:
I can provide a link to this page in the commit message if this is desired: https://www.erlang.org/doc/design_principles/applications.html#directory-structure-guidelines-for-a-development-environment 

In general `-` are not allowed in application names, as it's no a proper atom. Tooling like rebar3 has already evolved to cope with such limitations by providing aliases and other workarounds, but naming apps directories this way is not a very good practice. It also lead to extra complexity when I tried to set it up with the mix/rebar3 mixed project which originally motivated me to change that. 

In the current state of Antidote it probably does not cause any immediate problems so should be viewed as a kind of a cleanup.

PR #484, opened 2022-06-09
Lock plugin versions and OTP upgrade

albsch:
Remove support for older OTP versions, linting requires OTP 23

* Workflow uses Erlang 23 & 24
* Locked rebar3 plugins
* Formatted all files with `erlfmt`

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy: 
       matrix:
-        erlang: [ 21,22,23,24 ]
+        erlang: [ 23,24 ]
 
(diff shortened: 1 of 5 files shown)


define-null:
You might want to lock plugin versions as well, that was the cause for build break. We did the following in our fork:
(https://github.com/vaxine-io/vaxine/blob/5e206664346d2acfbd35f2c0e72dd1943112789f/rebar.config#L6)
By the way rebar3_format is probably redundant, given that erlfmt is used at the moment.

albsch:
`rebar3_format` is needed to be able to use `rebar3 format` without calling `erlfmt` directly.

bieniusa (approved):

PR #488, opened 2022-06-15
Fix ordering issue in vector_orddict

define-null:
In the current master `vector_orddict:insert` function would compare all dots for `>`, rather then `>=`, which will lead to a wrong ordering of vv, such as: if given
`#{dc1 => 1, dc2 => 1}` and `#{dc1 => 1, dc=> 2}`, the vector_orddict would be dependent on the insert order, and vector_orddict:first would give different results.


bieniusa:
Thanks for spotting the issue!
It seems that the function is not used, so this wasn't creating any problem.
The actually used `insert_bigger` employs `vectorclock:le\2` so this should be correct. 

I suggest removing `all_dots_greater` and `all_dots_smaller` as they are not used anywhere and the generic `all_dots` is more useful and less ambiguous, might also consider removing the unused functions from `vector_orddict`.

PR #499, opened 2026-04-13
erlang 28 fixes

albsch:
* maybe is reserved and ignore some warnings
* removed non-existant options
* fixed some internal types
* Bumped minimal supported Erlang version to 27
* Newer Erlang versions make the partitions tests fail because of overlapping partitions (A,B, runner nodes are connected, B disconnects from A, but not from runner -> overlapping partitions)
* Partitioning cluster never really working in the common test suite, because of overlapping partitions and race conditions. Partition tests should not be flaky anymore. Using `peer` module now (required for Erlang 29)


apps/antidote/src/antidote_pb_process.erl
```diff
@@ -48,7 +48,7 @@ from_bin(Clock) ->
 encode_clock(TxId) ->
     term_to_binary(TxId).
 
--spec process(antidote_pb_codec:request()) -> antidote_pb_codec:response_in().
```
albsch, line 51: dangling ref

albsch (commented):
self-review

PR #500, opened 2026-04-14
Unique module names and improved workflow

albsch:
Module names have to be unique. Added reltest to workflow.


albsch (commented):
self-review

albsch (commented):
self review

